Issue #23, opened 2025-01-31
list index out of range error using DeepSeek


ThomasCosyn:
Hello,

I tried to integrate a DeepSeek model deployed on Azure with AzureAIChatCompletionsModel, but I get the error 
```
File site-packages\langchain_azure_ai\chat_models\inference.py", line 500, in _stream  
choice = chunk.choices[0]  
IndexError : list index out of range
```

I guess the API sometimes sends a response with a `choices `empty list. We could check the size of the list before getting the first item ?

```    def _stream(
        self,
        messages: List[BaseMessage],
        stop: Optional[List[str]] = None,
        run_manager: Optional[CallbackManagerForLLMRun] = None,
        **kwargs: Any,
    ) -> Iterator[ChatGenerationChunk]:
        inference_messages = to_inference_message(messages)
        default_chunk_class = AIMessageChunk
        response = self._client.complete(
            messages=inference_messages,
            stream=True,
            stop=stop or self.stop,
            **self._identifying_params,
            **kwargs,
        )
        for chunk in response:
            if len(chunk.choices) > 0:
                choice = chunk.choices[0]
                chunk = _convert_delta_to_message_chunk(choice.delta, default_chunk_class)
                finish_reason = choice.finish_reason
                generation_info = (
                    dict(finish_reason=finish_reason) if finish_reason is not None else None
                )
                default_chunk_class = chunk.__class__  # type: ignore[assignment]
                cg_chunk = ChatGenerationChunk(
                    message=chunk, generation_info=generation_info
                )
                if run_manager:
                    run_manager.on_llm_new_token(cg_chunk.text, chunk=cg_chunk)
                yield cg_chunk```

efriis:
cc @marlenezw

marlenezw:
Thanks for the issue, just merged a fix from @patcheng 🙏🏾

PR #25, opened 2025-02-05
Fixed index out of range and missing usage data

patcheng:
I was trying to fix https://github.com/langchain-ai/langchain-azure/issues/23

I also noticed that usage information was not being filled in.

Looks like `_generate` methods looks very similar and was populating the the usage information, so I take the code from `_generate` and bolted it into `_stream`/`_astream`

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/chat_models/inference.py
+++ b/libs/azure-ai/langchain_azure_ai/chat_models/inference.py
@@ -7,6 +7,7 @@ from typing import (
     AsyncIterator,
     Callable,
     Dict,
+    Iterable,
     Iterator,
     List,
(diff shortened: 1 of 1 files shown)


marlenezw:
Thanks for the PR @patcheng , I'm reviewing!

marlenezw:
Awesome! Thank you Patrick this works as expected. I'm merging 
cc @santiagxf

schatekar:
@marlenezw  when do you think you will be able to release the new version? I desperately need this fix as it prevents me from pushing something into our QA environment.

marlenezw:
Hi @schatekar, we're working on getting the release out by the end of today!

PR #29, opened 2025-02-06
Bug fix for, error when inserting more than 500 documents at once

anushakolan:
### What is this change about?

There is a bug in the current implementation of **add_texts**, where if we insert more than 500 texts or documents at once into the DB, it throws an exception as below.
```
2024-12-02 11:02:35,662 - ERROR - Add text failed:
 ('07002', '[07002] [Microsoft][ODBC Driver 18 for SQL Server]COUNT field incorrect or syntax error (0) (SQLExecDirectW)')
```
### How is it fixed?

Use batching to insert the documents into the database based on the batch_sie value.
1. The batch_size is an optional parameter for inserting documents / data.
2. The default batch_size will be 100.

### How are the changes tested?

Integration tests were added to insert 636 documents at once and also 1000 texts at once using **add_texts** API. All  the tests were successful.
Screenshot 2025-02-06 002906

### DIFF
--- a/libs/sqlserver/langchain_sqlserver/vectorstores.py
+++ b/libs/sqlserver/langchain_sqlserver/vectorstores.py
@@ -813,6 +813,7 @@ class SQLServer_VectorStore(VectorStore):
         texts: Iterable[str],
         metadatas: Optional[List[dict]] = None,
         ids: Optional[List[str]] = None,
+        batch_size: Optional[int] = 100,
         **kwargs: Any,
     ) -> List[str]:
         """`add_texts` function for SQLServer_VectorStore class.
@@ -824,16 +825,33 @@ class SQLServer_VectorStore(VectorStore):
             texts: Iterable of strings to add into the vectorstore.
             metadatas: List of metadatas (python dicts) associated with the input texts.
             ids: List of IDs for the input texts.
+            batch_size: Number of documents to be inserted at once to Db, max 500
             **kwargs: vectorstore specific parameters.
 
         Returns:
             List of IDs generated from adding the texts into the vectorstore.
         """
-        # Embed the texts passed in.
-        embedded_texts = self.embedding_function.embed_documents(list(texts))
+        if texts is None:
+            return []
 
-        # Insert the embedded texts in the vector store table.
-        return self._insert_embeddings(texts, embedded_texts, metadatas, ids)
+        # Initialize a list to store results from each batch
+        embedded_texts = []
+
+        # Loop through the list of documents and process in batches
+
+        for i in range(0, len(list(texts)), batch_size):
+            batch = texts[i : i + batch_size]
+            batch_ids = ids[i : i + batch_size] if ids is not None else None
+            batch_metadatas = (
+                metadatas[i : i + batch_size] if metadatas is not None else None
+            )
+            batch_result = self.embedding_function.embed_documents(list(batch))
+            embeddings = self._insert_embeddings(
+                batch, batch_result, batch_metadatas, batch_ids
+            )
+            embedded_texts.extend(embeddings)
+
+        return embedded_texts
 
     def drop(self) -> None:
         """Drops every table created during initialization of vector store."""
@@ -1184,6 +1202,7 @@ class SQLServer_VectorStore(VectorStore):
             if ids is None:
                 # Get IDs from metadata if available.
                 ids = [metadata.get("id", uuid.uuid4()) for metadata in metadatas]
+                print("nnn ", ids)
 
             with Session(self._bind) as session:
                 documents = []
--- a/libs/sqlserver/langchain_sqlserver/vectorstores.py
+++ b/libs/sqlserver/langchain_sqlserver/vectorstores.py
@@ -1202,7 +1202,6 @@ class SQLServer_VectorStore(VectorStore):
             if ids is None:
                 # Get IDs from metadata if available.
                 ids = [metadata.get("id", uuid.uuid4()) for metadata in metadatas]
-                print("nnn ", ids)
 
             with Session(self._bind) as session:
                 documents = []
--- a/libs/sqlserver/tests/integration_tests/test_vectorstores.py
+++ b/libs/sqlserver/tests/integration_tests/test_vectorstores.py
@@ -7,6 +7,7 @@ from unittest.mock import Mock
 
 import pytest
 from langchain_core.documents import Document
+from langchain_text_splitters import RecursiveCharacterTextSplitter
 from sqlalchemy import create_engine, text
 
 from langchain_sqlserver.vectorstores import DistanceStrategy, SQLServer_VectorStore
@@ -859,3 +860,36 @@ def connect_to_vector_store(conn_string: str) -> SQLServer_VectorStore:
         embedding_function=DeterministicFakeEmbedding(size=EMBEDDING_LENGTH),
         table_name=_TABLE_NAME,
     )
+
+
+def test_sqlserver_batch_add_documents(
+    store: SQLServer_VectorStore,
+    texts: str,
+) -> None:
+    """Test that `add_documents` returns equivalent number of ids of input
+    texts when using more than 500 documents."""
+
+    text_splitter = RecursiveCharacterTextSplitter(chunk_size=3, chunk_overlap=1)
+    split_documents = text_splitter.create_documents(texts)
+
+    result = store.add_documents(split_documents)
+    assert len(result) == len(split_documents)
+
+
+def test_sqlserver_batch_add_texts(
+    store: SQLServer_VectorStore,
+    texts: str,
+) -> None:
+    """Test that `add_texts` returns equivalent number of ids of input
+    texts when using more than 500 texts."""
+    texts *= 200
+    result = store.add_texts(texts)
+    assert len(result) == len(texts)
+
+
+def test_sqlserver_batch_add_texts_no_texts(
+    store: SQLServer_VectorStore,
+) -> None:
+    """Test that `add_texts` returns 0 ids when no texts"""
+    result = store.add_texts(None)
+    assert len(result) == 0


libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -813,6 +813,7 @@ def add_texts(
         texts: Iterable[str],
         metadatas: Optional[List[dict]] = None,
         ids: Optional[List[str]] = None,
+        batch_size: Optional[int] = 100,
```
beccadaniel: If we introduce this here, it means that users might not be able to pass `batch_size` if say `add_documents` is called. They'd have to default to a `batch_size` of 100.

It might be a better idea to introduce `batch_size` as a class member and be able to use it in the entire class.

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -824,16 +825,34 @@ def add_texts(
...
-        embedded_texts = self.embedding_function.embed_documents(list(texts))
+        if texts is None:
+            return []
+
+        # Initialize a list to store results from each batch
+        embedded_texts = []
+
+        # Loop through the list of documents and process in batches
+        texts = list(texts)
+        batch_size = batch_size if batch_size is not None else 100
```
beccadaniel: Can we have this default value defined as a constant for readability?

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -859,3 +860,36 @@ def connect_to_vector_store(conn_string: str) -> SQLServer_VectorStore:
...
+
+
+def test_sqlserver_batch_add_documents(
+    store: SQLServer_VectorStore,
+    texts: str,
+) -> None:
+    """Test that `add_documents` returns equivalent number of ids of input
+    texts when using more than 500 documents."""
+
+    text_splitter = RecursiveCharacterTextSplitter(chunk_size=3, chunk_overlap=1)
```
beccadaniel, line 864: Just curious. Does this end up creating a large number of documents considering we're starting a small sized text?

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: Okay

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: Yeah, because the chunk size is 3, it is creating 525 documents.

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: `def add_documents(self, documents: list[Document], **kwargs: Any) -> list[str]:,
`
So my thought was, similar to how ids are added below,

```
if "ids" not in kwargs:
                ids = [doc.id for doc in documents]

                # If there's at least one valid ID, we'll assume that IDs
                # should be used.
                if any(ids):
                    kwargs["ids"] = ids
```
users can send batch_size as a kwargs argument.

anushakolan: `def add_documents(self, documents: list[Document], **kwargs: Any) -> list[str]:, 
`So my thought was, similar to how ids are added below,
 
```
        if type(self).add_texts != VectorStore.add_texts:
            if "ids" not in kwargs:
                ids = [doc.id for doc in documents]

                # If there's at least one valid ID, we'll assume that IDs
                # should be used.
                if any(ids):
                    kwargs["ids"] = ids

            texts = [doc.page_content for doc in documents]
            metadatas = [doc.metadata for doc in documents]
            return self.add_texts(texts, metadatas, **kwargs)
```
 
 
users can send batch_size as a **kwargs argument.

anushakolan: Updated the code

anushakolan: If we go with **kwargs approach, each insert function in the future needs to be separately updated with the batch_size when needed. So declared it as a class variable instead.

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -215,6 +217,8 @@ def __init__(
                 Optional param, defaults to None.
             table_name: The name of the table to use for storing embeddings.
                 Default value is `sqlserver_vectorstore`.
+            batch_size: Number of documents/texts to be inserted at once to Db, max 500.
```
beccadaniel: We don't enforce the max of 500, should we be doing that?

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -824,16 +835,36 @@ def add_texts(
             texts: Iterable of strings to add into the vectorstore.
             metadatas: List of metadatas (python dicts) associated with the input texts.
             ids: List of IDs for the input texts.
+            batch_size: Number of documents to be inserted at once to Db, max 500.
```
beccadaniel: This documentation is not needed here.

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: Yeah, if we don't enforce 500, we will anyway get the error. So, let's wait for Davide's and Ani's reply, will update the code based on that.

anushakolan: Updating to 419 as per the offline discussion.

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: Updating to 419 as per the offline discussion.

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -215,21 +217,38 @@ def __init__(
...
             connection if connection else self._create_engine()
         )
         self._prepare_json_data_type()
         self._embedding_store = self._get_embedding_store(self.table_name, self.schema)
         self._create_table_if_not_exists()
 
+    def _validate_batch_size(self, batch_size: int) -> int:
+        if batch_size is None or batch_size <= 0:
+            return DEFAULT_BATCH_SIZE
+        elif batch_size > 419:
```
Aniruddh25: Define 419 as a const - MAX_BATCH_SIZE

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -215,21 +217,38 @@ def __init__(
...
+        self._batch_size = batch_size
         self._bind: Union[Connection, Engine] = (
             connection if connection else self._create_engine()
         )
         self._prepare_json_data_type()
         self._embedding_store = self._get_embedding_store(self.table_name, self.schema)
         self._create_table_if_not_exists()
 
+    def _validate_batch_size(self, batch_size: int) -> int:
+        if batch_size is None or batch_size <= 0:
```
Aniruddh25: Shouldn't a -ve or 0 batch_size be invalid value as well? Shouldn't we error on that value? - with a different error message than the maximum one

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -824,16 +849,36 @@ def add_texts(
             texts: Iterable of strings to add into the vectorstore.
             metadatas: List of metadatas (python dicts) associated with the input texts.
             ids: List of IDs for the input texts.
+            batch_size: Number of documents to be inserted at once to Db, max 419.
```
beccadaniel: This is no longer an arg for `add_texts`. We should remove this.

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -824,16 +849,36 @@ def add_texts(
             texts: Iterable of strings to add into the vectorstore.
```
Aniruddh25: Where is the batch_size argument added to add_texts? If this function doesn't take it as an argument, the PR description should be fixed, and so does the argument list

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -215,21 +217,38 @@ def __init__(
...
         self._embedding_store = self._get_embedding_store(self.table_name, self.schema)
         self._create_table_if_not_exists()
 
+    def _validate_batch_size(self, batch_size: int) -> int:
+        if batch_size is None or batch_size <= 0:
+            return DEFAULT_BATCH_SIZE
+        elif batch_size > 419:
+            logging.error("The request contains an invalid batch_size.")
+            raise ValueError(
+                """The request contains an invalid batch_size.
```
beccadaniel: Maybe we can include the actual value passed in by the user in the error message too.

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -824,16 +849,36 @@ def add_texts(
...
+            return []
+
+        # Initialize a list to store results from each batch
+        embedded_texts = []
+
+        # Loop through the list of documents and process in batches
+        texts = list(texts)
+
+        # Validate batch_size again to confirm if it is still valid.
+        batch_size = self._validate_batch_size(self._batch_size)
```
Aniruddh25, line 875: why would it not be valid if it was already validated?

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -824,16 +849,36 @@ def add_texts(
...
+        for i in range(0, len(list(texts)), batch_size):
```
Aniruddh25: isnt texts already a list(texts) as per 865? If yes, shouldnt this be 0, len(texts) ?

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -824,16 +849,36 @@ def add_texts(
...
         """
-        # Embed the texts passed in.
-        embedded_texts = self.embedding_function.embed_documents(list(texts))
+        if texts is None:
+            return []
+
+        # Initialize a list to store results from each batch
+        embedded_texts = []
+
+        # Loop through the list of documents and process in batches
```
Aniruddh25: Is it a list of texts or documents? If texts why do we say documents here in the comment? If documents, why is the variable named texts?

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -824,16 +849,36 @@ def add_texts(
...
+        texts = list(texts)
+
+        # Validate batch_size again to confirm if it is still valid.
+        batch_size = self._validate_batch_size(self._batch_size)
+        for i in range(0, len(list(texts)), batch_size):
+            batch = texts[i : i + batch_size]
```
Aniruddh25, line 877: What if texts is < batch_size, does this indexing mechanism account for that? and is this excluding the last value `i + batch_size`?

libs/sqlserver/poetry.lock
```diff
@@ -13,35 +13,35 @@ files = [
 
 [[package]]
 name = "anyio"
-version = "4.6.2.post1"
+version = "4.8.0"
```
Aniruddh25, line 16: what are these changes? and are they required?

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -850,6 +851,82 @@ def test_that_connection_string_with_trusted_connection_yes_does_not_use_entra_i
...
+def test_sqlserver_batch_add_documents_with_batch_size(
+    store: SQLServer_VectorStore,
+    texts: List[str],
+) -> None:
+    """Test that `add_documents` returns equivalent number of ids of input
+    texts when using more than 500 documents and custom batch_size."""
+
+    text_splitter = RecursiveCharacterTextSplitter(chunk_size=3, chunk_overlap=1)
+    split_documents = text_splitter.create_documents(texts)
+    store._batch_size = 400
```
Aniruddh25: the test is modifying the batch_size private member directly. This should not be how we test it. It should be tested via the interface how batch_size is exposed to the user. Is it as an argument to add_documents, add_texts, or from_documents or from_texts API?

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -850,6 +851,82 @@ def test_that_connection_string_with_trusted_connection_yes_does_not_use_entra_i
...
+
+
+def test_sqlserver_batch_add_texts_raises_exception(
+    store: SQLServer_VectorStore,
+    texts: List[str],
+) -> None:
+    """Test that `add_texts` raises an exception,
+    when batch_size is updated to more than 419"""
+    texts *= 200
+    store._batch_size = 490
```
Aniruddh25: Need to add a test with -ve value of batch_size

Aniruddh25 (changes requested):
Need more accurate testing

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: Batch_size is an optional parameter, if none is mentioned we go for the default batch_size, otherwise there will be no use for the default batch_size.

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: Defined constant and re-used it.

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: Yes, the batch_size argument will not be added to individual functions. It is added only to the constructor. Will update the PR description to make this clear.

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: The intension here is add_douments() in the base.py internally calls add_texts with the texts extracted from the document. So, I added the comment loop through the list of documents. But we actually are looping through the texts.
Tthe comment is misleading. Updating the comment to "list of texts"

libs/sqlserver/poetry.lock
anushakolan: This change was auto generated. We added a new dependency langchain-text-splitters, the peotry.lock file could be changed because of some dependency chain created by langchain-text-splitters.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: 1. batch_size will not be a parameter to add_documents, add_texts API. This can only be added via the object instantiation.
2. from_documents or from_texts are class methods, that are used to instantiate the class object, hence they do support these parameters.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: Yes, I will add a testcase for the -ve value as well.

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: This is done to satisfy the build tool. As the tool complains about potential possibility of the _batch_size value being updated after initializing. I am adding steps like,
1. naming the variable name with _, so that it indicates private variable. 
2. Creating a property attribute for the variable batch_size.
But python does not enforce editing this private variable, so there is nothing much we can do about it.
There will still be a possibility to edit the variable using class name as mentioned in this link. To avoid that and to confirm that the variable is still valid we check it again. https://stackoverflow.com/questions/62986995/how-to-prevent-class-attribute-from-being-modified-from-outside-class

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: Yes, added a test to verify that as well. Yes, it excludes` i + batch_size`.

libs/sqlserver/langchain_sqlserver/vectorstores.py
Aniruddh25: You are right about the value None. We should go for the default batch_size when it is None. My question was for the `batch_size <= 0` condition, shouldnt we error for that case?

anushakolan: This is a choice Ani, if the user gives an invalid batch_size, we can log and error and continue with batching with default batch_size because the rest of the data is valid.

But if we want to be strict about batch_size we should error out. Let me bring this discussion with Davide.

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -215,21 +218,38 @@ def __init__(
...
         """
+        batch_size = self._validate_batch_size(batch_size)
         self.connection_string = self._get_connection_url(connection_string)
         self._distance_strategy = distance_strategy
         self.embedding_function = embedding_function
         self._embedding_length = embedding_length
         self.schema = db_schema
         self.override_relevance_score_fn = relevance_score_fn
         self.table_name = table_name
+        self.__batch_size = batch_size
```
beccadaniel: nit: for consistency should `__batch_size` be `_batch_size`. Python docs say double leading and trailing underscores are reserved for special use in the language. This is supposed to be a private variable.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -868,55 +869,56 @@ def test_sqlserver_batch_add_documents(
...
+    """Test that when store is re-initialized with a different batch_size,
+    the new value is not accepted by the store object. In below case we should not get
+    an error even with a new batch_size(900) > 419, because that value cannot be
+    assigned after instantiating the object."""
 
     text_splitter = RecursiveCharacterTextSplitter(chunk_size=3, chunk_overlap=1)
     split_documents = text_splitter.create_documents(texts)
-    store._batch_size = 400
+    store = connect_to_vector_store(_CONNECTION_STRING_WITH_UID_AND_PWD, 230)
+    store._batch_size = 900
```
Aniruddh25: In C++, this would have been a compile error. I know Python is an interpreted language, no compilation - so wouldnt this end up being a runtime error since _batch_size shouldnt be accessible?

Aniruddh25: Or is the private member named `__batch_size` and since this is `_batch_size` notice the difference in name, does it mean we are setting a new member altogether?

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -868,55 +869,56 @@ def test_sqlserver_batch_add_documents(
...
+def test_sqlserver_batch_add_documents_with_negative_batch_size_uses_default_batch_size(
     texts: List[str],
 ) -> None:
-    """Test that `add_texts` raises an exception,
-    when batch_size is updated to more than 419"""
-    texts *= 200
-    store._batch_size = 490
-    with pytest.raises(ValueError):
-        store.add_texts(texts)
+    """Test that when a store is initialized with a negative batch_size,
```
Aniruddh25: Based on the comment above, we need this to throw an error.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -868,55 +869,56 @@ def test_sqlserver_batch_add_documents(
...
-    store._batch_size = 0
-    result = store.add_texts(texts)
-    assert len(result) == len(texts)
+    """Test that when a store is initialized with a batch_size less than texts size,
+    it will not throw an exception."""
+
+    # creates 33 documents, len(texts) = 33, batch_size = 400
+    text_splitter = RecursiveCharacterTextSplitter(chunk_size=50, chunk_overlap=0)
+    split_documents = text_splitter.create_documents(texts)
+    store = connect_to_vector_store(_CONNECTION_STRING_WITH_UID_AND_PWD, 400)
```
Aniruddh25: nit: name the argument 400 for easier readability.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -868,55 +869,56 @@ def test_sqlserver_batch_add_documents(
...
+def test_sqlserver_batch_add_documents_with_texts_less_than_batch_size(
     texts: List[str],
 ) -> None:
-    """Test that `add_texts` returns equivalent number of ids of input
-    texts when using more than 500 texts."""
-    texts *= 200
-    store._batch_size = 0
-    result = store.add_texts(texts)
-    assert len(result) == len(texts)
+    """Test that when a store is initialized with a batch_size less than texts size,
```
Aniruddh25: nit: this comment doesnt match the values set in the test. len(texts) = 33 and batch_size=400.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: It is just used once for test, do we still want to name it and use it?

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: I did not get this Ani. I added len(text) =33 to highlight that 33 documents are created. batch_size used is 400.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: I did not change the test yet in this revision, so the comment is still the same. Updated both the test and the comment in the next revision.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -29,9 +29,9 @@
     texts as filter_texts,
 )
 
-pytest.skip(
-    "Skipping these tests pending resource availability", allow_module_level=True
-)
+# pytest.skip(
```
Aniruddh25, line 32: Why is this commented? If not needed, better to remove it from the file

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -459,6 +484,7 @@ def from_texts(
         db_schema: Optional[str] = None,
         distance_strategy: DistanceStrategy = DEFAULT_DISTANCE_STRATEGY,
         ids: Optional[List[str]] = None,
+        batch_size: int = DEFAULT_BATCH_SIZE,
```
Aniruddh25, line 487: I thought we agreed that none of the APIs will get this additional argument and only the constructor will initialize it. Am I missing something?

libs/sqlserver/tests/integration_tests/test_vectorstores.py
Aniruddh25: I mean when you call the function, if python allows: `connect_to_vector_store(_CONNECTION_STRING_WITH_UID_AND_PWD, batch_size: 400)`

libs/sqlserver/tests/integration_tests/test_vectorstores.py
Aniruddh25: if len(text) = 33 and batch_size = 400, is batch_size less than texts size? No, right? Im highlighting that the comment is wrong. should be:

Test that when a store is initialized with a batch_size **greater** than texts size,

Aniruddh25 (changes requested):
Overall looks good. Still waiting on confirmation where would the user specify batch_size - just the constructor or the individual APIs like from_texts, from_documents

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: This should not be commented; I had commented it as part of testing. Will remove it.

libs/sqlserver/langchain_sqlserver/vectorstores.py
anushakolan: Both these methods are @classmethods, which instantiate the SQLServer_VectorStore class. Since, we are instantiating the SQL instance here, I had to add the parameter batch_size to these methods. Rest are API's which do not instantiate the SQLServer_VectorStore class, instead they use the store object and call the APIs.

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: Sending this fix in new revision

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: Will fix in new revision

Aniruddh25 (approved):
Thanks for addressing all concerns!

beccadaniel (approved):
Looks good. 👍🏼!

PR #34, opened 2025-02-09
Adding some fixex for full text and hybrid search

aayush3011:


santiagxf:
@aayush3011 can we host sample notebooks somewhere else? Not sure if we are hosting examples in this repository. @marlenezw what do you think? Otherwise we would need to host a more complehensive documentation and samples.

marlenezw:
Yes, I think for the moment let's continue to host our docs and sample notebooks in the main Langchain repo so they show on the Langchain website as expected.  Ayush, I think you could make updates here: https://github.com/langchain-ai/langchain/tree/master/docs/docs/integrations/vectorstores 

We could maybe have a docs folder here in this repo and host some examples, but those would be specifically for contributors to use.

Issue #35, opened 2025-02-10
Generated tool calls should have a `type` key, requests with messages with tool calls without `type` returning a `424` error


charbeltabet:
### Description
When comparing the generated tool calls between `ChatOpenAI` and `AzureAIChatCompletionsModel`, we see that the tool calls of the azure integration **lack the `'type'` key**. In the case of open ai, it has `'type': 'function'`.

This is a problem because sending such a tool call back without type to azure in a messages array returns a 424 error response:

```
Message: Missing required parameter: 'messages[2].tool_calls[0].type'.
```

<img width="829" alt="Image" src="https://github.com/user-attachments/assets/bade0d18-bf13-4228-a9d3-2d711a0d3278" />

### Steps to reproduce
Note that both are using the same `gpt-4o-mini-2024-07-18` model

```py
import os
from langchain_azure_ai.chat_models import AzureAIChatCompletionsModel
from langchain_openai import ChatOpenAI
from langchain_core.tools import tool
from langchain_core.messages import HumanMessage

llm = AzureAIChatCompletionsModel(
    endpoint=AZURE_ENDPOINT,
    credential=AZURE_CREDENTIAL,
    model_name="gpt-4o-mini",
)

OPENAI_API_KEY = os.getenv("OPENAI_API_KEY")
openai_llm = ChatOpenAI(model="gpt-4o-mini")

@tool
def dummy_tool(query: str) -> str:
  """
    Dummy tool for testing purposes

    Args:
      query: The query to return

    Returns:
      str: The query
  """
  return query

dummy_tools = [
  dummy_tool
]
azure_llm_with_tools = llm.bind_tools(dummy_tools)
openai_llm_with_tools = openai_llm.bind_tools(dummy_tools)

azure_tool_calls = azure_llm_with_tools.invoke([HumanMessage(content="Test the dummy tool")])
openai_tool_calls = openai_llm_with_tools.invoke([HumanMessage(content="Test the dummy tool")])

print("Azure tool calls:", azure_tool_calls)
print("OpenAI tool calls:", openai_tool_calls)

[...]

Issue #19, opened 2025-01-17
content are missing in message


Amitgb14:
Sometimes we noticed the content are missing from message,
https://github.com/langchain-ai/langchain-azure/blob/main/libs/azure-ai/langchain_azure_ai/chat_models/inference.py#L140

`{'role': 'assistant', 'tool_calls': None}`

marlenezw:
Thanks, I'll run some tests with this. cc @santiagxf

santiagxf:
Thanks for reaching out @Amitgb14. Do you mind sharing a bit more details about the code you are running so we are sure we got it right? We are shipping some fixes related to the use of tools already in version 0.1.1. I want to make sure we captured your case.

Amitgb14:
Hello @santiagxf , here are some details,

```
{ "message": "unexpected the task, err: err=2 validation errors for AIMessage\ncontent.str\n Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]\n For further information visit https://errors.pydantic.dev/2.9/v/string_type\ncontent.list[union[str,dict[any,any]]]\n Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]\n For further information visit https://errors.pydantic.dev/2.9/v/list_type" }
```
and it comes only for Azure AI Non GPT models.

```config

        "inference_endpoint":"https://myendpoint.services.ai.azure.com/models",
        "model_id": "Ministral-3B",
```

And I can't share input here.

santiagxf:
@Amitgb14 unfortunately I'm not able to tell what you are doing. I don't need to know the inputs you are passing, but please tell me which parameters are you using. I'm assuming you are using the class `AzureAIChatCompletionsModel` and you are creating a chain with it? Are you using tools? or any specific feature?

Amitgb14:
@santiagxf , I'm using ```AzureAIChatCompletionsModel```,  we seen this are frequent issue

```
from langchain_azure_ai.chat_models import AzureAIChatCompletionsModel


model = AzureAIChatCompletionsModel(
                endpoint=inference_endpoint,
                model_name=self.model_id if self.model_id else None,
                credential=api_key,
                api_version=api_version,
                http_async_client=httpx.AsyncClient(proxy=proxy) if proxy else None,
                **kwargs)
```

santiagxf:
[...]

PR #36, opened 2025-02-10
Add type when formatting tool call

charbeltabet:
This addresses #35 by adding the `type` of the `tool_call` which is _always function for function calls_ according to [OpenAI's documentation](https://platform.openai.com/docs/guides/function-calling#streaming).

Not having this field results in a `424` error response from azure described in this [issue](https://github.com/langchain-ai/langchain-azure/issues/35).

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/chat_models/inference.py
+++ b/libs/azure-ai/langchain_azure_ai/chat_models/inference.py
@@ -259,7 +259,8 @@ def _format_tool_call_for_azure_inference(tool_call: ToolCall) -> dict:
         "function": {
             "name": tool_call["name"],
             "arguments": json.dumps(tool_call["args"]),
-        }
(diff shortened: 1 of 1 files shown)


marlenezw:
Thanks for the PR @charbeltabet, will test and merge. This also might help with issue #19.

santiagxf (approved):

santiagxf:
Thanks @charbeltabet for the great contribution! LGTM! We are taking you suggestion and I'm adding some tests to prevent this from happening again. Will release a patch as soon as we get test coverage.

charbeltabet:
How should we approach test this @santiagxf ? I have this PR #38 where I add a test for this, let me know what you think. Maybe we shouldn't test private functions and there is a better way of testing this? Maybe by editing a mock or something

PR #52, opened 2025-02-27
Pass the model name to each request for routing

patcheng:
According to https://learn.microsoft.com/en-us/azure/ai-foundry/model-inference/concepts/endpoints?tabs=python#routing

`complete` takes a `model` parameter that is needed for routing.

Use the model_name, and pass it along.


patcheng:
I found that model parameter was already being passed it. My error was caused by something else. Closing this PR

PR #58, opened 2025-03-25
Fix: Update User Agent for vCore

SelinaTang21:
Update User Agent to `langchainpy` for consistency in vCore tracking


marlenezw:
LGTM! Great work @SelinaTang21!

PR #64, opened 2025-04-03
tracing: external Azure Monitor config

ianchi:
Forcing the configuration of Azure Monitor inside an instrumentation is not a natural pattern.

This doesn't allow any customization of the configuration (like passing the logger name or credentials).

It also doesn't work well if you already had previously configured the monitor for other instrumentations (for instance FastApi in my case).

This PR sepparates concerns, as is the commont pattern in all opentelemetry instrumentations.
It doesn't add much complexity to the user and it is easy if well documented.


ianchi:
Hi @santiagxf, any comments on this?

ianchi:
Friendly ping

ianchi:
Hi @santiagxf, any chance to review this?

Please let me know if I can help somehow

santiagxf:
We made `connection_string` optional in #85. let me know if this works for you!

PR #66, opened 2025-04-05
Update pyproject.toml to remove unnecessary dependencies

marlenezw:
PR to remove:
- numpy
- pymongo 
- simsimd 

These do not seem like necessary dependencies and for some reason, pip is struggling when it hits these. Will test but P0 to remove whatever is causing the slow install


marlenezw:
Creating a PR for this so we can discuss whats causing the slow pip install @santiagxf . My feeling is that these could be the issue, but need to test

marlenezw:
Yeah, when I just ran the pip install it's breaking at pymongo. Need to remove and publish a new release 

`WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='pypi.org', port=443): Read timed out. (read timeout=15)")': /simple/pymongo/`

marlenezw:
Merging and will wait for Facundo to release the updated package. This should significantly improve install time I think. Right now it is too slow. 

Notes:
-added numpy, pymongo and simsimd to linting and testing dependencies. 
- Cosmos team should put a note in their docs or that people should install pymongo when they want to use the mongodb vector store cc @aayush3011 
- simsimd is used for cosine similarity in utils but errors if not installs and tells the user to install it so should be fine. Its not needed as a main dependency 
- numpy is useful but very tricky to have as a main dependency because it breaks things if its the wrong version for a package. Most packages will install the version of the package when they use it

For linting the packages should be installed automatically but let me know if you start seeing errors @santiagxf

Issue #65, opened 2025-04-03
bug: error in open telemetry tracer


ianchi:
When using  AzureAIInferenceTracer I get constant errors:

```
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x7f4f12ad5300> at 0x7f4ec83a1fc0> was created in a different Context
ERROR:  opentelemetry.context Failed to detach context
```

This is triggered in the `_end_span` function when calling `context_api.detach`

The code runs inside a FastApi endpoint handler.

```python
# Initialization
tracing_handler = AzureAIInferenceTracer(
      connection_string=settings.external.APPLICATIONINSIGHTS_CONNECTION_STRING,
      enable_content_recording=True,
  )

# Usage
@router.post("/stream")
async def stream( .....):
   # .....
   graph.astream(
                state,
                {
                    "configurable": {"thread_id": data.thread_id},
                    "callbacks": [tracing_handler],
                })

```

santiagxf:
Hi @ianchi for reaching out about this issue. Give us some time to reproduce the problem and get back to you shortly.

ianchi:
Thanks @santiagxf 

It is not directly related to this error, but another problem we are facing when trying to use this tracing is described in #64

ianchi:
Hi @santiagxf ,

Appart from the original error that is still present, I also see errors when the state/input/output is a pydantic object and not a dict:

```
Error in tracing callback: Object of type Send is not JSON serializable
```

Another case:
```
Error in tracing callback: Object of type State is not JSON serializable
```

In all cases the ofending line is:

- code.filepath: /home/apiuser/.local/lib/python3.11/site-packages/langchain_azure_ai/callbacks/tracers/inference_tracing.py 
- code.function: _handle_event_error 
- code.lineno: 155

ianchi:
Hi @santiagxf, any comment about this?
For us tracing in Azure is completely broken due to this issue.

santiagxf:
@ianchi can you check if this was fixed on the latest [release](https://github.com/langchain-ai/langchain-azure/actions/runs/15183711860)?

ianchi:
Hi @santiagxf, thanks! 2 out of 3 problems are solved (pydantic objects and instrumentation setup).
But the original one in this issue (first post) is still there.

We still get flooded by these errors:
```
ERROR:  opentelemetry.context Failed to detach context
Traceback (most recent call last):
[...]

PR #68, opened 2025-04-08
Fix: handle pydantic in inference tracing

ianchi:
Solves second part of #65


ianchi:
Hi, any comments about this?

ianchi:
Hi @marlenezw,

Any chance to merge this?
We still cannot use inference tracing.

Issue #70, opened 2025-04-09
Python 3.13 needs Numpy > 2.0


ACPAleksR:
If you attempt to install langchain-azure, which needs numpy=^1.26.4 you will get issues on Python 3.13, which needs
numpy>=2.0.0.

Error Message and Stack Trace (if applicable)
https://github.com/langchain-ai/langchain/issues/26026
https://github.com/numpy/numpy/issues/24318

Description
The numpy dependency needs bumping up to >=2.0.0

System Info
Python 3.13 on Windows 11

marlenezw:
Thanks for the issue! We will address this in the next release.

sydney-runkle:
Fixed this on the `langchain-core` end of things via https://github.com/langchain-ai/langchain/pull/30714 :)

marlenezw:
Thank you @sydney-runkle!!!

sydney-runkle:
Sure thing. Should be released now in `langchain-core` v0.3.52 :)

szatixx:
@marlenezw I am still having the numpy issue while attempting to install langchain-azure-ai package. 

I have python 3.13.3 and this packages: 

aiohappyeyeballs                                  2.6.1
aiohttp                                           3.11.18
aiosignal                                         1.3.2
annotated-types                                   0.7.0
anthropic                                         0.50.0
anyio                                             4.9.0
attrs                                             25.3.0
azure-core                                        1.33.0
azure-identity                                    1.21.0
azure-monitor-opentelemetry-exporter              1.0.0b36
backoff                                           2.2.1
certifi                                           2025.1.31
cffi                                              1.17.1
charset-normalizer                                3.4.1
colorama                                          0.4.6
cryptography                                      44.0.2
dataclasses-json                                  0.6.7
Deprecated                                        1.2.18
distro                                            1.9.0
filelock                                          3.18.0
fixedint                                          0.1.6
frozenlist                                        1.6.0
fsspec                                            2025.3.2
googleapis-common-protos                          1.70.0
greenlet                                          3.2.1
grpcio                                            1.71.0
h11                                               0.14.0
httpcore                                          1.0.8
httpx                                             0.28.1
httpx-sse                                         0.4.0
huggingface-hub                                   0.30.2
idna                                              3.10
importlib_metadata                                8.6.1
inflection                                        0.5.1
isodate                                           0.7.2
Jinja2                                            3.1.6
jiter                                             0.9.0
jsonpatch                                         1.33
[...]

PR #72, opened 2025-04-11
Updating dependencies to include correct versions of numpy

marlenezw:
Adding numpy back as a dependency because I see it is needed for most things. Just updating the versions to be appropriate with the Python version


marlenezw:
Should close [issue](https://github.com/langchain-ai/langchain-azure/issues/70)

Issue #12, opened 2024-12-02
[langchain-sqlserver] Generated UUID are truncated


yorek:
When saving a document into SQL Server vectorstore without providing an ID, a UUID is generated automatically for the document, but the generated value is saved with some truncation. This is the UUID I see in my database:

0A565240-1B3B-4D

anushakolan:
Looks like the `id` is not truncated, but the `custom_id` is truncated.

anushakolan:
PR for bug fix, Pull Request #73.

PR #73, opened 2025-04-18
Bug fix: Fix UUIDs truncation for custom_id

anushakolan:
**Why is this change needed?**

This change addresses the bug, https://github.com/langchain-ai/langchain-azure/issues/12.
The issue here is when no `custom_id` is provided by the user while inserting a text or document, the python langchain package generates a UUID and stores it in the DB. This `custom_id` value is being truncated.
This change fixes the bug, so that the `custom_id` is not truncated.

**What is reason for the bug?**

The bug is because, the `custom_id` column is of type `varchar(1000) `. The UUID is of type `UniqueIdentifier`. 

When you use `uuid.uuid()` in SQL Server, it generates a UUID object. If you directly insert this object into a varchar column, SQL Server must implicitly convert the UUID object to a string. Depending on the SQL Server version and the settings, this implicit conversion might not always convert the UUID to its string representation correctly, potentially leading to truncation or other unexpected behavior.

**What is the fix for the bug?**

To fix this bug, the UUID is explicitly casted to str and then stored in the DB. 

**How are the changes tested?**

A new test case is added to verify that the UUIDs for `custom_id` are not truncated. This test does the following.

1. Inserts a single text and its metadata into the database without specifying an ID using the add_texts API.
This allows the python langchain package to generate and use random UUIDs.
2. Uses the returned IDs from the add_texts API to retrieve the inserted document information via the get_by_ids API.
3. Extracts the custom_id value from the document information.
4. Compares the length of the custom_id to the length of a UUID to ensure that the UUID has not been truncated.

Also, all the existing tests are run to ensure these changes does not introduce any new bugs.


libs/sqlserver/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "langchain-sqlserver"
-version = "0.1.1"
+version = "0.1.2"
```
Aniruddh25, line 3: Do we plan to create the new patch after this bug fix is checked in?

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -1243,9 +1243,9 @@ def _insert_embeddings(
                     # For a query, if there is no corresponding ID,
                     # we generate a uuid and add it to the list of IDs to be returned.
                     if idx < len(ids):
-                        custom_id = ids[idx]
+                        custom_id = str(ids[idx]).upper()
                     else:
-                        ids.append(str(uuid.uuid4()))
+                        ids.append(str(uuid.uuid4()).upper())
```
Aniruddh25: Looks like it is in python that we generate the UUID if its not provided. See here. uuid.uuid4() 
In description you mention it is the system:

`, the system generates a UUID and stores it in the DB. `

Please be specific - `the python langchain package generates a UUID and stores it in DB`

libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -1243,9 +1243,9 @@ def _insert_embeddings(
                     # For a query, if there is no corresponding ID,
                     # we generate a uuid and add it to the list of IDs to be returned.
                     if idx < len(ids):
-                        custom_id = ids[idx]
+                        custom_id = str(ids[idx]).upper()
```
Aniruddh25: Instead of doing conversion of uuid to str, do we really need to keep the datatype of `id` column to be `varchar(1000)` ? Should we instead change the datatype to uuid? 

What are the consequences of doing so?

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -925,6 +925,30 @@ def test_sqlserver_batch_add_documents_with_batch_size_edited(
...
+    result = store.add_texts(texts, metadatas)
+
+    # Use the ids from the add_texts to get the inserted document data.
+    documents = store.get_by_ids(result)
+    print(documents)
+    custom_id = documents[0].id
+
+    if custom_id is not None:
+        # Verify the length of the custom_id is equal to UUID length
+        assert len(custom_id) == uuid_len
```
Aniruddh25: can we provide an error message if the test fails?

libs/sqlserver/tests/integration_tests/test_vectorstores.py
```diff
@@ -925,6 +925,30 @@ def test_sqlserver_batch_add_documents_with_batch_size_edited(
...
+    store: SQLServer_VectorStore,
+) -> None:
+    """Test that triggers UUID creation and
+    verifies that the value is stored in the db without truncation."""
+
+    # Insert a text with no id, so that UUIDs are generated and used.
+    texts = ["hello"]
+    metadatas = [{"summary": "Good Quality Dog Food"}]
+    uuid_len = 36
+    result = store.add_texts(texts, metadatas)
```
Aniruddh25, line 938: Description says that 

>This allows the database to generate and use random UUIDs.

However, it is NOT the database who generates the random UUIDs, we generate it in python client then send it to the database. Please fix description.

libs/sqlserver/langchain_sqlserver/vectorstores.py
yorek: The custom_id is set by the end user, and they can use anything as an id. It could be a UUID but also something else. So I think we should keep the varchar(1000) as the data type

libs/sqlserver/pyproject.toml
anushakolan: Yes, that is the plan

libs/sqlserver/tests/integration_tests/test_vectorstores.py
anushakolan: Okay

beccadaniel:
>When a UUID is casted to str, it is by default converted to lower case. Since the primary key ID is directly stored as UUID, it is in upper case. To maintain consistency with the primary key, the custom_id is also converted to upper case and stored.

How / where are we doing this conversion to uppercase?

Aniruddh25:
> When a UUID is casted to str, it is by default converted to lower case. Since the primary key ID is directly stored as UUID, it is in upper case. To maintain consistency with the primary key, the `custom_id` is also converted to upper case and stored.

This needs to be removed from description since we are no longer upper casing it

Aniruddh25 (approved):

anushakolan:
> > When a UUID is casted to str, it is by default converted to lower case. Since the primary key ID is directly stored as UUID, it is in upper case. To maintain consistency with the primary key, the custom_id is also converted to upper case and stored.
> 
> How / where are we doing this conversion to uppercase?

We should not, removed it.

beccadaniel (approved):

PR #55, opened 2025-03-08
add support for Azure CosmosDB NoSQL native WHERE and query_constructors

ianchi:
This is a resubmission of #44 against the current release branch.

All description available in that PR.

PR #77, opened 2025-05-03
AzureCosmosDBNoSqlVectorSearchl: Improvements & fixes

ianchi:
Some additional enhancements to AzureCosmosDBNoSqlVectorSearch:

- Fix translator when using bool values
- currently the database is created even if paramater says you don't want to. Fix that and make parameters optional
- Add true async support (with optional async client)

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/vectorstores/azure_cosmos_db_no_sql.py
+++ b/libs/azure-ai/langchain_azure_ai/vectorstores/azure_cosmos_db_no_sql.py
@@ -50,10 +50,10 @@ class AzureCosmosDBNoSqlVectorSearch(VectorStore):
         *,
         cosmos_client: CosmosClient,
         embedding: Embeddings,
-        vector_embedding_policy: Dict[str, Any],
-        indexing_policy: Dict[str, Any],
-        cosmos_container_properties: Dict[str, Any],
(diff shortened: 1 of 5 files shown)


ianchi:
Hi @marlenezw / @aayush3011 , this builds on top of #55 and adds some more enhancements.

I've grouped three of them in this PR, in separate commits,

ianchi:
Hi, any comments on this?

libs/azure-ai/langchain_azure_ai/vectorstores/azure_cosmos_db_no_sql.py
```diff
@@ -460,6 +521,64 @@ def similarity_search_with_score(
             )
         return docs_and_scores
 
+    async def asimilarity_search_with_score(
```
aayush3011, line 524: Add `asimiliarity_search` method as well.

aayush3011 (commented):
Left a comment, rest of the PR looks good.

libs/azure-ai/langchain_azure_ai/vectorstores/azure_cosmos_db_no_sql.py
ianchi: It is already there. Row 490

PR #96, opened 2025-06-09
Creating langchain-azure-storage

marlenezw:
This PR creates a new `langchain-azure-storage` package


libs/azure-storage/langchain_azure_storage/azure_blob_storage_file.py
```diff
@@ -0,0 +1,45 @@
...
+            raise ImportError(
+                "Could not import azure storage blob python package. "
+                "Please install it with `pip install azure-storage-blob`."
+            ) from exc
+
+        client = BlobClient.from_connection_string(
+            conn_str=self.conn_str, container_name=self.container, blob_name=self.blob
+        )
+
+        with tempfile.TemporaryDirectory() as temp_dir:
```
tonybaloney, line 38: Use `NamedTemporaryFile` instead and do-away with the `file_path` and the `os.makedirs` call. You can just write directly to a temporary file,

https://docs.python.org/3.6/library/tempfile.html#tempfile.NamedTemporaryFile

This also handles deleting the file for you afterward, which the current code doesnt and would leave the file in tmp

marlenezw: The storage team should take over this PR and can make these changes. cc @kyleknap. We should also do a security review once you've made changes to the branch

kyleknap: Yeah we are still doing research on the interfaces/behaviors we want here. I'm not even sure we want to be writing to disk at all here; writing to disk can slow down the transfer speed and there are no guarantees that it will have the permissions to write to disk. Instead, I'd like to explore streaming in-memory for this.

We chatted offline that we are going to build the storage integrations as part of a feature branch. I created the tracking feature branch under [azure-storage/main](https://github.com/langchain-ai/langchain-azure/tree/azure-storage/main) with the intent that once the integration is ready to be released, we'd merge it into main. I also switched the target of this PR to that feature branch and can use the PR to initialize the langchain-azure-storage package there.

marlenezw: Awesome! This all sounds good to me!

marlenezw:
Going to close this PR since we are using another branch for working the new package

PR #111, opened 2025-06-16
langchain_azure_storage setup

anjaliratnam-msft:
Removed the files in the langchain_azure_storage folder.


libs/azure-storage/tests/.gitignore
```diff
@@ -0,0 +1,3 @@
+__pycache__
```
kyleknap: We should probably hoist the `.gitignore` to the `libs/azusre-storage` directory so it applies to all python related files. This is also similar to how the other azure langchain libraries are set up.

libs/azure-storage/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 LangChain, Inc.
```
kyleknap: Something we should probably want to double check with @marlenezw and @santiagxf here... Do we want LangChain on the license even though this library is being authored by Microsoft? All of the other azure langchain libraries use the same copyright but seemed like something worth asking given other Microsoft authored libraries (e.g. in the Azure GitHub org) use Microsoft in the license.

libs/azure-storage/README.md
```diff
@@ -0,0 +1,9 @@
+# langchain-azure-storage
+
+This package contains the LangChain integration for Azure Blob and Container Storage.
```
kyleknap: A couple of suggestions here:
* Let's make integration plural e.g. `integrations` instead of `integration`
* Let's update the ending from `Azure Blob and Container Storage` to `Azure Blob Storage` which is the official service name

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
+[tool.poetry]
+name = "azure-storage"
+version = "0.1.0"
+description = "Azure Storage library for LangChain Azure"
```
kyleknap: Let's update the description to `Azure Storage integrations for LangChain`. Mainly that feels a bit more crisp.

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+authors = []
```
kyleknap: For the authors, lets go ahead and copy what we did for the PyTorch connector: https://github.com/Azure/azure-storage-for-pytorch/blob/476d8eedfd7ea925366194b12bc11f88c782d70f/pyproject.toml#L7-L9. This also matches the authors used in the Azure Python SDK.

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+license = "MIT"
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = ">=3.9,<4.0"
+langchain-core = "^0.3.0"
+typing-extensions = "^4.9.0"
+unstructured = "^0.8.0"
+azure-storage-blob = "^12.16.0"
+azure-identity = "^1.16.0"
```
kyleknap: Let's remove all of these runtime dependencies for now. We can add them back when we implement features. After removing them, we will probably need to regenerate the poetry lock.

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+[tool.mypy]
+disallow_untyped_defs = "True"
+
+# Suppress missing stubs warnings for unstructured and its submodules
+[[tool.mypy.overrides]]
+module = [
+    "unstructured",
+    "unstructured.*"
+]
+ignore_missing_imports = true
```
kyleknap: Let's remove this section for now since we will be removing `unstructured` as a dependency.

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+codespell = "^2.2.0"
+
+[tool.poetry.group.lint.dependencies]
+ruff = "^0.5"
+python-dotenv = "^1.0.1"
+pytest = "^7.4.3"
+tomli = "^2.2.1"
+
+[tool.poetry.urls]
+"Source Code" = "https://github.com/langchain-ai/langchain-azure/tree/main/libs/sqlserver"
```
kyleknap: We should update this url to point to the `azure-storage` library instead of the `sqlserver`

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
+[tool.poetry]
+name = "azure-storage"
+version = "0.1.0"
```
kyleknap: It would be interesting to see if we can look into how to be able to declare the package version once (either in the `pyproject.toml` or `__version__.py`) and the value be propagated across the package instead of needing to maintain the version in both places. It would be worth looking to see if `poetry` has the notion of dynamic metadata values similar to setuptools: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata which we have leveraged in other projects.

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+python = ">=3.9,<4.0"
+langchain-core = "^0.3.0"
+typing-extensions = "^4.9.0"
+unstructured = "^0.8.0"
+azure-storage-blob = "^12.16.0"
+azure-identity = "^1.16.0"
+
+[tool.poetry.group.test.dependencies]
+pydantic = "^2.9.2"
+pytest = "^7.4.3"
```
kyleknap: Let's try to cut down the number of dependencies that we have in each of these groups for now unless we know that we need them. For this one, let's try removing everything but `pytest` and see if CI passes.

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+pytest-mock = "^3.10.0"
+pytest-watcher = "^0.3.4"
+pytest-asyncio = "^0.21.1"
+python-dotenv = "^1.0.1"
+syrupy = "^4.7.2"
+langchain-core = {git = "https://github.com/langchain-ai/langchain.git", subdirectory = "libs/core"}
+
+[tool.poetry.group.test_integration.dependencies]
+pytest = "^7.3.0"
+python-dotenv = "^1.0.1"
```
kyleknap: We should be able to remove the dependency on `python-dotenv` here. We can add it back if we decide to set env vars using it.

libs/azure-storage/pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+
+[tool.poetry.group.codespell.dependencies]
+codespell = "^2.2.0"
+
+[tool.poetry.group.lint.dependencies]
+ruff = "^0.5"
+python-dotenv = "^1.0.1"
```
kyleknap: We should be able to remove `python-dotenv` and I think `tomli` here. We should try to remove it and see if CI passes.

kyleknap (commented):
Looks good. I just had some smaller comments going through this initial setup.

libs/azure-storage/pyproject.toml
anjaliratnam-msft: It looks like for the poetry version being used, it doesn't allow the dynamic version line. It's only in the [2.1 version](https://python-poetry.org/docs/1.8/pyproject/). There's also the poetry-dynamic-versioning plugin but it looks like that's only for getting the version from git tags.

libs/azure-storage/LICENSE
marlenezw: This code came from the LangChain community library, so this is why it has LangChain, we could put LangChain and Microsoft, but we should leave some LangChain ownership in there

kyleknap: @marlenezw thanks for the context. That makes sense on why LangChain is included. I like the idea on including Microsoft as well especially given that we expect to be the primary maintainers and will likely be adding new features/integrations outside of what was originally contributed to the LangChain community project.

kyleknap: Went with the approach that @marlenezw suggested by adding both LangChain and Microsoft to the license. Resolving thread.

libs/azure-storage/pyproject.toml
kyleknap: Switched this to `Langchain integrations for Azure Storage`. This is more consistent with the README and other Azure integration packages.

libs/azure-storage/pyproject.toml
kyleknap: I updated the `__init__.py` to follow what the Azure AI package does to be able to set the version in a single place: https://github.com/langchain-ai/langchain-azure/blob/main/libs/azure-ai/langchain_azure_ai/__init__.py. This is what the main `langchain` package does and is what seems to be the generally recommended pattern when using poetry and trying to consolidate places were version is set without any additional dependencies.

marlenezw:
Ah! Awesome work @anjaliratnam-msft  and @kyleknap! Let me know if you want to plan a call to walk through the release workflow

PR #133, opened 2025-08-07
Update Poetry to the v2 Family

aytekinar:
Added support for Poetry v2 by

- [x] updating the `pyproject.toml` files of the libraries, and,
- [x] changing the `POETRY_VERSION` environment variable in the CI pipelines.

Because Poetry v2 has deprecated most of the fields in favour of [PEP 517][pep-517] and [PEP 621][pep-621], I have moved most of the `[tool.poetry.XXX]` stuff to their correct/new places in `pyproject.toml` files. This has also enabled me/us to have `uv` support inside the same `pyproject.toml` files, without conflicting with Poetry. However, not to get in the way of other developers, I have kept both Poetry and `uv` support in the `pyproject.toml` files (so that we can switch to `uv` when/if needed, or ditch that altogether, otherwise).

[pep-517]: https://peps.python.org/pep-0517/
[pep-621]: https://peps.python.org/pep-0621/

> [!NOTE]  
> I have done my best to transform the old caret requirements to the [PEP 440][pep-440]-style tilde/compatibility requirements and verified that `poetry check` gives an OK for the changes. That said, it may be worthwhile to let the tests run and even do some manual checks to see if there is something wrong about the packages.

[pep-440]: https://peps.python.org/pep-0440/

## Rationale

Poetry, in its v2 family of releases, has chosen to deprecate the so-called "caret requirements," which was not recognized by other Python packaging tools. Similarly, they have changed the way "tilde (compatibility) requirements" behave, i.e., they fixed the wrong behavior of v1 in their v2 family of releases (cf. the documentation for [`v1`][poetry-v1] and [`v2`][poetry-v2], and compare them against [PEP 440][pep-440]).

[poetry-v1]: https://python-poetry.org/docs/1.8/dependency-specification/#tilde-requirements
[poetry-v2]: https://python-poetry.org/docs/dependency-specification/#tilde-requirements


marlenezw:
Awesome work Arda! I run tests and then we can go from there! cc @santiagxf and @tonybaloney This PR will move us to poetry v2 and also make things work with uv!

marlenezw (approved):
LGTM!

PR #135, opened 2025-08-14
Propagate PGHOST and PGPASSWORD secrets to release tests

aytekinar:


aytekinar:
Auto-merging without approval because this is needed to fix the release test pipeline(s).

Apologies for going fast...

PR #149, opened 2025-09-17
Added dependencies for document loaders

anjaliratnam-msft:
Added the dependencies needed for the Azure Storage document loaders.


libs/azure-storage/pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 
-dependencies = []
```
kyleknap, line 25: For the new dependencies, let's do the following:

* Have each dependency on its own line instead of all on one line. That is the general pattern when declaring dependencies
* Use [compatible release](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release) notation. This is consistent with the other azure langchain packages in terms of pinning dependencies. For example, [this](https://github.com/langchain-ai/langchain-azure/blob/main/libs/azure-ai/pyproject.toml) is what `langchain-azure-ai` does.
* Let's also include the async dependencies for both `azure-identity` and `azure-storage-blob`. This will require instead declaring these dependencies as `azure-storage-blob[aio]` and `azure-identity[aio]`. This ensures we pull in any of the async dependencies required.
* Let's add a `dev` dependency group, that specifies the git repository `langchain-core` (e.g., like [this](https://github.com/langchain-ai/langchain-azure/blob/cbab74035757e4eabe4f43ea19f80a91f845f0a0/libs/azure-ai/pyproject.toml#L70)). This will allow us to test against the latest developments of langchain when developing this library.

kyleknap (commented):
Looks fine. Just had some suggestions.

libs/azure-storage/pyproject.toml
anjaliratnam-msft: For azure-identity[aio], I got a warning that aio is not an extra, so I think it's included and the aiohttp package is also already there.

kyleknap: Good catch. It looks like our SDK only uses the `aio` extra and not the identity package (e.g., here's the [`pyproject.toml`](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/pyproject.toml))

libs/azure-storage/pyproject.toml
```diff
@@ -22,7 +22,11 @@ classifiers = [
...
+dependencies = [
+    "azure-identity~=1.25.0",
+    "azure-storage-blob[aio]~=12.26.0",
+    "langchain-core~=0.3.76",
```
kyleknap: Let's set the langchain dependency to `~=0.3.0`. Mainly that matches what the azure langchain libraries seem to require and it gives customers more flexibility to pin to version in the 0.3.x range.

libs/azure-storage/pyproject.toml
```diff
@@ -45,6 +49,9 @@ optional = true
 [tool.poetry.group.codespell.dependencies]
 codespell = "~2.2"
 
+[tool.poetry.group.dev.dependencies]
```
kyleknap, line 55: In including the git repository of langchain for the `langchain-core` pacakge, we should also:

* Make sure to mark the `dev` group as optional such as [here](https://github.com/langchain-ai/langchain-azure/blob/cbab74035757e4eabe4f43ea19f80a91f845f0a0/libs/azure-ai/pyproject.toml#L64)
* Add the langchain repository for `langchain-core` as a source for `uv` such as [here](https://github.com/langchain-ai/langchain-azure/blob/cbab74035757e4eabe4f43ea19f80a91f845f0a0/libs/azure-ai/pyproject.toml#L154)
* Declare the `dev` group in the dependency group section like [here](https://github.com/langchain-ai/langchain-azure/blob/cbab74035757e4eabe4f43ea19f80a91f845f0a0/libs/azure-ai/pyproject.toml#L130)

libs/azure-storage/pyproject.toml
```diff
@@ -22,7 +22,11 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 
-dependencies = []
+dependencies = [
+    "azure-identity~=1.25.0",
+    "azure-storage-blob[aio]~=12.26.0",
```
kyleknap: Sorry small thing that I missed... We should remove the `0` patch version off of the azure sdk dependencies. So (`azure-identity~=1.25` and `azure-storage-blob[aio]~=12.26`). Mainly as is, we are pinning to a specific minor version but that prevents customers from upgrading in the next minor version release. Removing the `0` allows customers to upgraded automatically up to the next major version of the SDKs (e.g. a `2.*.*` version)

PR #150, opened 2025-09-23
AzureBlobStorageLoader lazy_load() implementation

anjaliratnam-msft:
Added the AzureBlobStorageLoader class and implemented the lazy_load method. Also added unit tests.

### DIFF
--- a/libs/azure-storage/langchain_azure_storage/document_loaders.py
+++ b/libs/azure-storage/langchain_azure_storage/document_loaders.py
@@ -74,6 +74,7 @@ class AzureBlobStorageLoader(BaseLoader):
             "account_url": self._account_url,
             "container_name": self._container_name,
             "credential": credential,
+            "connection_data_block_size": self._CONNECTION_DATA_BLOCK_SIZE,
         }
 
     def _lazy_load_documents_from_blob(
--- a/libs/azure-storage/tests/unit_tests/test_document_loaders.py
+++ b/libs/azure-storage/tests/unit_tests/test_document_loaders.py
@@ -193,7 +193,7 @@ def test_async_credential_provided_to_sync(
 def test_invalid_credential_type(
     create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
 ) -> None:
-    mock_credential = ""
+    mock_credential = "account-key"
     with pytest.raises(TypeError, match="Invalid credential type provided."):
         create_azure_blob_storage_loader(
             blob_names="text_file.txt", credential=mock_credential


libs/azure-storage/langchain_azure_storage/__init__.py
```diff
@@ -2,6 +2,9 @@
 
 from importlib import metadata
 
+from .document_loaders import AzureBlobStorageLoader
+
+__all__ = ["AzureBlobStorageLoader"]
```
kyleknap: Let's hold off importing the document loader into the root of the package. While it simplifies customer imports especially since we only have a single langchain abstraction, if we expand to other langchain components, importing to the `__init__` can complicate imports and slow down import of the package. Instead, we can add this later if we see customers wanting to be able to import from the root of the package.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
+"""LangChain integrations for Azure Storage."""
```
kyleknap: We should probably make this doc string specific to document loaders.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+import azure.core.credentials
+import azure.core.credentials_async
+import azure.identity
+from azure.storage.blob import BlobClient, ContainerClient
+from langchain_core.document_loaders import BaseLoader
+from langchain_core.documents.base import Blob, Document
+
+SDK_CREDENTIAL_TYPE = Optional[
+    Union[
+        azure.identity.DefaultAzureCredential,
```
kyleknap: Do we need to declare `DefaultAzureCredential` as a type here? It should be a subclass of the `TokenCredential` class.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+        self._credential = credential
+        self._loader_factory = loader_factory
+
+    def lazy_load(self) -> Iterator[Document]:
+        """Lazily load documents from the specified blobs in the container."""
+        self._credential = _get_sdk_credential(self._credential)
+        if isinstance(
+            self._credential, azure.core.credentials_async.AsyncTokenCredential
+        ):
+            raise ValueError("Async credential provided to sync method.")
```
kyleknap: Do we need this additional check here? Would it be handled by the `_get_sdk_credential` call?

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+        account_url: str,
+        container_name: str,
+        blob_names: Optional[Union[str, Iterable[str]]] = None,
+        prefix: str = "",
+        credential: Optional[
+            Union[
+                azure.core.credentials.AzureSasCredential,
+                azure.core.credentials.TokenCredential,
+            ]
+        ] = None,
```
kyleknap: Can we use the SDK credential type that we defined previously?

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class AzureBlobStorageLoader(BaseLoader):
+    """Document loader for loading LangChain Document objects from a container."""
+
+    def __init__(
+        self,
+        account_url: str,
+        container_name: str,
+        blob_names: Optional[Union[str, Iterable[str]]] = None,
```
kyleknap, line 30: Let's add an `*` between the `blob_name` and `prefix` parameters to enforce that starting with `prefix` it needs to be provided by keyword argument. More details here on the feature: https://peps.python.org/pep-3102/ but by forcing these arguments to be keyword only, it gives us more flexibility in the ordering of some of these auxiliary parameters as they would not be able to be specified positionally.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+        prefix: str = "",
```
kyleknap: Let's make `prefix` an `Optional[str]`. It's sort of nuanced but it will allow us to validate whether the customer provided a prefix and blob name as we can check whether the prefix is not `None` as opposed to falsey.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+            azure.core.credentials.TokenCredential,
+            azure.core.credentials.AzureSasCredential,
+        ),
+    ):
+        return credential
+    raise TypeError("Invalid credential type provided.")
+
+
+class AzureBlobStorageLoader(BaseLoader):
+    """Document loader for loading LangChain Document objects from a container."""
```
kyleknap: Let's update this docstring to say it loads documents from Azure Blob Storage instead of a container since we consolidated the two clases.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+
+def _lazy_load_documents_from_blob(blob_client: BlobClient) -> Iterator[Document]:
+    blob_data = blob_client.download_blob(max_concurrency=10)
+    blob = Blob.from_data(
+        data=blob_data.readall(),
+        mime_type=blob_data.properties.content_settings.content_type,
+        metadata={
+            "source": blob_client.url,
+        },
+    )
```
kyleknap: I realize this is from the prototype that I originally written, but we should probably remove the logic of turning it into a `Blob` object and just directly instantiate the `Document` to simplify the implementation as we as of now not exposing a `blob_parser` parameter.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+SDK_CREDENTIAL_TYPE = Optional[
+    Union[
+        azure.identity.DefaultAzureCredential,
+        azure.core.credentials.AzureSasCredential,
+        azure.core.credentials.TokenCredential,
+    ]
+]
+
+
+def _get_client_kwargs(credential: SDK_CREDENTIAL_TYPE = None) -> dict:
```
kyleknap: Let's move all of these module level helpers to private methods on the `AzureBlobStorage` class especially since we no longer have two classes and it will help improve readability by hoisting the main class higher in this file.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+        container_client = ContainerClient.from_container_url(
+            container_url,
+            **_get_client_kwargs(self._credential),
+        )
+        blobs_with_prefix = list(
+            container_client.list_blob_names(name_starts_with=self._prefix)
+        )
+        if self._blob_names:
+            for blob_name in self._blob_names:
+                if blob_name in blobs_with_prefix:
```
kyleknap: It would be interesting to encapsulate this logic into a helper method `_yield_blob_names()` that encapsulates both yielding from a listing blobs in the container and from the iterator. Mainly this will help remove some of the amount of nesting that is currently needed in the main method.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+
+def _get_sdk_credential(credential: SDK_CREDENTIAL_TYPE) -> SDK_CREDENTIAL_TYPE:
+    if credential is None:
+        return azure.identity.DefaultAzureCredential()
+    if isinstance(
+        credential,
+        (
+            azure.core.credentials.TokenCredential,
+            azure.core.credentials.AzureSasCredential,
+        ),
```
kyleknap: Would it be possible to reuse the `_SDK_CREDENTIAL_TYPE` that we previously declared to use for credential type validation.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+from typing import Callable, Iterable, Iterator, Optional, Union
+
+import azure.core.credentials
+import azure.core.credentials_async
+import azure.identity
+from azure.storage.blob import BlobClient, ContainerClient
+from langchain_core.document_loaders import BaseLoader
+from langchain_core.documents.base import Blob, Document
+
+SDK_CREDENTIAL_TYPE = Optional[
```
kyleknap: We should make this type start with an underscore (`_SDK_CREDENTIAL_TYPE`) so that it does not appear as part of the public interface and possibly be imported.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,112 @@
...
+            Union[
+                azure.core.credentials.AzureSasCredential,
+                azure.core.credentials.TokenCredential,
+            ]
+        ] = None,
+        loader_factory: Optional[Callable[[str], BaseLoader]] = None,
+    ):
+        """Initialize AzureBlobStorageLoader.
+
+        Parameters:
```
kyleknap: Let's update these to follow [Google Style of Python docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings). Poking around the langchain and langchain-azure repository, it looks like it follows that style. So, it would make sense to be consistent here. For example, [here](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/document_loaders/base.py) is how langchain documents the base loaders. Also, I think we should be able to omit the types in the docstrings since we already have type annotations on them.

kyleknap (commented):
It's looking good. Overall, it was just smaller comments on the implementation. I did not get to looking at the tests; I'll do that in my next pass.

libs/azure-storage/langchain_azure_storage/document_loaders.py
anjaliratnam-msft: I think isinstance only takes in tuples so that doesn't work, but I could use get_args(_SDK_CREDENTIAL_TYPE) if that's better.

kyleknap: Yeah let's try using `get_args()` for now especially since we will be effectively referring to the same set of classes in the validation

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+        Args:
+            account_url: URL to the Azure Storage account, e.g. "https://<account_name>.blob.core.windows.net"
+            container_name: Name of the container to retrieve blobs from in the
+                storage account
+            blob_names: List of blob names to load. If None, all blobs will be loaded
+            prefix: Prefix to filter blobs in the container
+            credential: Credential to authenticate with the Azure Storage account.
+                If None, DefaultAzureCredential will be used
+            loader_factory: A callable that returns a custom document loader to use for
+                parsing blobs downloaded. Ex: PyPDFLoader, CSVLoader, etc.
```
kyleknap: Let's go ahead and hold off adding the `loader_factory` to the signature until we actually implement it. Mainly that will help cut down on the diff and I think we probably want to expand on this docstring a bit more on how the mechanics work.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+            blob_list = list(container_client.list_blob_names())
+            for blob_name in blob_list:
+                if blob_name in self._blob_names:
+                    yield blob_name
+        elif self._prefix:
+            yield from container_client.list_blob_names(name_starts_with=self._prefix)
+        else:
+            yield from container_client.list_blob_names()
+
+    def lazy_load(self) -> Iterator[Document]:
```
kyleknap: Nit-pick but let's hoist this method just before the `__init__`. Mainly it is a little easier to read through the class when the public methods come before the private methods.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+        """Lazily load documents from the specified blobs in the container.
```
kyleknap: Let's just make this docstring `Lazily loads documents from Azure Blob Storage` as the loader could be set up to not have specific blobs configured to load.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+
+        Yields:
+            the documents.
+        """
+        self._credential = self._get_sdk_credential(self._credential, "sync")
+        container_url = f"{self._account_url}/{self._container_name}"
+        container_client = ContainerClient.from_container_url(
+            container_url,
+            **self._get_client_kwargs(self._credential),
+        )
```
kyleknap: Because the `account_url` and `container_name` are provided as separate arguments, it probably makes sense to just using the constructor on the `ContainerClient` and delegate any URI building logic to the SDK instead of using `from_container_url()`.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+            if function_type == "sync" and isinstance(
+                credential, azure.core.credentials_async.AsyncTokenCredential
+            ):
+                raise TypeError("Async credential provided to sync method.")
+            return credential
+        raise TypeError("Invalid credential type provided.")
+
+    def _yield_blob_names(self, container_client: ContainerClient) -> Iterator[str]:
+        if self._blob_names is not None and self._prefix is not None:
+            raise ValueError("Cannot specify both blob_names and prefix.")
```
kyleknap: We should probably validate this at loader construction time instead of `lazy_load()` runtime. Mainly from a general interface perspective, it is expected that any misconfigurations are raised as early as possible. Validation during runtime method invocation should really only be considered if it is not possible otherwise to do as part of the constructor.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if self._blob_names:
+            blob_list = list(container_client.list_blob_names())
+            for blob_name in blob_list:
+                if blob_name in self._blob_names:
+                    yield blob_name
```
kyleknap: Ah so if `blob_names` are provided, we don't want to be making any list API calls. Mainly one of the advantages in providing the blob names yourself is that you no longer need to make the hop to determine what blobs are to be loaded and save on making N API calls for listing. I think we should be able to simplify this down to:
```python
yield from self._blob_names
```

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+        elif self._prefix:
+            yield from container_client.list_blob_names(name_starts_with=self._prefix)
+        else:
+            yield from container_client.list_blob_names()
```
kyleknap: Could we simplify this down to?
```python
else:
    yield from container_client.list_blob_names(name_starts_with=self._prefix)
```
Mainly if the SDK allows us to set `None` for the `name_starts_with` we can avoid the extra branching.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+        self, credential: _SDK_CREDENTIAL_TYPE, function_type: Literal["sync", "async"]
+    ) -> _SDK_CREDENTIAL_TYPE:
+        if credential is None:
+            return azure.identity.DefaultAzureCredential()
+        if isinstance(
+            credential,
+            (
+                azure.core.credentials.TokenCredential,
+                azure.core.credentials.AzureSasCredential,
+            ),
```
kyleknap: Similar to the `preview` and `blob_names` validation, we should validate provided credentials as part of the constructor to bubble up any known issues as soon as possible. Then because the check for using sync credentials with sync methods needs to be done at runtime, we can continue doing that check here.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+        ):
+            if function_type == "sync" and isinstance(
```
kyleknap: Instead of accepting a string parameter to determine the mode we are retrieving credentials, let's just make them explicit methods (e.g. `_get_sync_credential()` and eventually `_get_async_credential()`). Mainly, it is a bit safer from a type checking perspective and reduces the number of if branching that we'd need in the methods to make it easier to follow.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,113 @@
...
+        loader_factory: Optional[Callable[[str], BaseLoader]] = None,
+    ):
+        """Initialize AzureBlobStorageLoader.
+
+        Args:
+            account_url: URL to the Azure Storage account, e.g. "https://<account_name>.blob.core.windows.net"
+            container_name: Name of the container to retrieve blobs from in the
+                storage account
+            blob_names: List of blob names to load. If None, all blobs will be loaded
+            prefix: Prefix to filter blobs in the container
```
kyleknap: Let's expand on this docstring to indicate:
1. It is only applicable when listing blobs in the container
2. Cannot be used with `blob_names`.

kyleknap (commented):
This is looking good. Still focused most of the comments on the implementation so we are solid there before moving on to the tests.

libs/azure-storage/pyproject.toml
```diff
@@ -67,6 +67,7 @@ optional = true
 
 [tool.poetry.group.test.dependencies]
 pytest = "~7.4"
+pytest-mock = "~3.14"
```
kyleknap: Let's hold off pulling in `pytest-mock` and instead let's just use the `mock` module from `unittest`. Mainly I'm not sure if there is a strong reason we need this pytest plugin since `pytest-mock` is a slim wrapper over `mock` and it's advantageous to keep our dependency list as lean as possible.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,160 @@
...
+    list(loader.lazy_load())
+    mock_blob_client.assert_called_once_with("text_file.txt")
+
+
+def test_default_credential(mocker: MockerFixture) -> None:
+    mock_default_credential = mocker.patch.object(
+        azure.identity, "DefaultAzureCredential"
+    )
+    mock_default_credential.return_value = mocker.MagicMock()
+    mock_container_client = mocker.patch.object(ContainerClient, "from_container_url")
```
kyleknap: Instead of patching as part of the test, it might make sense to just apply an auto-fixture to patch the ContainerClient class such as we do [here in the PyTorch connector](https://github.com/Azure/azure-storage-for-pytorch/blob/40cd9b66eaa0a9214553dc7f4e69ebf813063e5e/tests/unit/test_client.py#L301-L315). Mainly that will guarantee that we never try to make API calls as part of the unit tests and also help reduce some of the verbosity of the test case bodies.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,160 @@
...
+    blob_contents: list[str],
+    mock_lazy_load_documents_from_blob: None,
+) -> None:
+    mock_container_client = mocker.patch.object(ContainerClient, "from_container_url")
+    mock_client = mocker.MagicMock()
+    mock_client.list_blob_names.return_value = blob_names
+    mock_container_client.return_value = mock_client
+
+    args = {**_blob_storage_loader_args(), "blob_names": blob_names}
+    loader = AzureBlobStorageLoader(**args)
```
kyleknap: It might be worth just making a `create_azure_blob_storage_loader()` pytest fixture. Sort of similar to what we do [here](https://github.com/Azure/azure-storage-for-pytorch/blob/40cd9b66eaa0a9214553dc7f4e69ebf813063e5e/tests/unit/test_datasets.py#L19) in the PyTorch Connector. Mainly it is advantageous because we can then expose things like `account_url` and `container_name` as fixtures and can then be directly asserted against instead of having to reduplicate these values.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,160 @@
...
+                page_content=content,
+                metadata={
+                    "source": f"https://testaccount.blob.core.windows.net/test-container/{name}"
+                },
+            )
+        )
+    mocker.patch(
+        "langchain_azure_storage.document_loaders.AzureBlobStorageLoader._lazy_load_documents_from_blob",
+        return_value=iter(docs),
+    )
```
kyleknap: Instead of patching the loader, we should instead be updating our mock container and blob clients to list and return the contents in the expected order. The main reason that I'm suggesting that is in bypassing the logic around downloading the blob contents and turning them into a document, we will have a gap in our tests as we technically would not be exercising that method any more since its been patched out.

kyleknap (commented):
Just leaving some general suggestions on how we have the tests set up as I just started looking through them

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,96 @@
...
+_SDK_CREDENTIAL_TYPE = Optional[
+    Union[
+        azure.core.credentials.AzureSasCredential,
+        azure.core.credentials.TokenCredential,
+    ]
+]
+
+
+class AzureBlobStorageLoader(BaseLoader):
+    """Document loader for loading LangChain Document object from Azure Blob Storage."""
```
kyleknap: Small nitpick... Let's make object plural here so `objects` instead of `object`.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,96 @@
...
+            blob_content.decode("utf-8"), metadata={"source": blob_client.url}
+        )
+
+    def _get_sync_credential(
+        self, credential: _SDK_CREDENTIAL_TYPE
+    ) -> _SDK_CREDENTIAL_TYPE:
+        if credential is None:
+            return azure.identity.DefaultAzureCredential()
+        if isinstance(credential, azure.core.credentials_async.AsyncTokenCredential):
+            raise TypeError("Async credential provided to sync method.")
```
kyleknap: For this exception, we should maybe expand the content to make it a little more clear what exactly happened and how to resolve because it may not be too straightforward on why this is happening. Maybe update it to:
```
Cannot use synchronous load methods when AzureBlobStorageLoader is instantiated using an AsyncTokenCredential. Use its asynchronous load methods instead or supply a synchronous TokenCredential to its credential parameter.
```

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,96 @@
...
+        yield Document(
+            blob_content.decode("utf-8"), metadata={"source": blob_client.url}
+        )
+
+    def _get_sync_credential(
+        self, credential: _SDK_CREDENTIAL_TYPE
+    ) -> _SDK_CREDENTIAL_TYPE:
+        if credential is None:
+            return azure.identity.DefaultAzureCredential()
+        if isinstance(credential, azure.core.credentials_async.AsyncTokenCredential):
```
kyleknap: Let's also make it a `ValueError` for this exception as technically async credentials will be accepted but it as a value will not supported for synchronous calls.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,96 @@
...
+        self._blob_names = [blob_names] if isinstance(blob_names, str) else blob_names
+        self._prefix = prefix
+
+        if credential is None or isinstance(credential, get_args(_SDK_CREDENTIAL_TYPE)):
+            self._credential = credential
+        else:
+            raise TypeError("Invalid credential type provided.")
+
+    def _get_client_kwargs(self, credential: _SDK_CREDENTIAL_TYPE = None) -> dict:
+        return {"credential": credential, "connection_data_block_size": 256 * 1024}
```
kyleknap: For the default configuration values, it may be worth defining them as a class level private variable (e.g., `_CONNECTION_DATA_BLOCK_SIZE` and `_MAX_CONCURRENCY`) as part of the document loader class to help with readability.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,96 @@
...
+        azure.core.credentials.AzureSasCredential,
+        azure.core.credentials.TokenCredential,
+    ]
+]
+
+
+class AzureBlobStorageLoader(BaseLoader):
+    """Document loader for loading LangChain Document object from Azure Blob Storage."""
+
+    def lazy_load(self) -> Iterator[Document]:
```
kyleknap: Whoops I misspoke here. I meant after the `__init__`. Generally `__init__` will go first and then the public methods after.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,96 @@
...
+        self, credential: _SDK_CREDENTIAL_TYPE
+    ) -> _SDK_CREDENTIAL_TYPE:
+        if credential is None:
+            return azure.identity.DefaultAzureCredential()
+        if isinstance(credential, azure.core.credentials_async.AsyncTokenCredential):
+            raise TypeError("Async credential provided to sync method.")
+        return credential
+
+    def _yield_blob_names(self, container_client: ContainerClient) -> Iterator[str]:
+        if self._blob_names:
```
kyleknap: Let's make this line `if self._blob_names is not None:`. Mainly this ensures we go down this branch if a value is provided to blob_names so if for example an empty list is provided, we don't end up listing the entire container when instead it should have resulted in just no documents being returned back.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+    return ["text_file.txt", "json_file.json", "csv_file.csv"]
+
+
+@pytest.fixture
+def blob_contents() -> list[str]:
+    return [
+        "test content",
+        "{'test': 'test content'}",
+        "col1,col2\nval1,val2",
+    ]
```
kyleknap, line 29: Would it make sense to combine these into a single `blobs` fixture to keep better track of which names match to which content and reduce the number of fixtures needed to be requested? I'd imagine we could make this a list of dictionaries e.g.
```python
[
   {"blob_name": "name", "blob_content": "content"},
   ...
]
```
and that may simplify some of the test bodies.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+def create_azure_blob_storage_loader(
+    account_url: str, container_name: str
+) -> Callable[..., AzureBlobStorageLoader]:
+    def _create_azure_blob_storage_loader(
+        account_url: str = account_url,
+        container_name: str = container_name,
+        blob_names: Optional[Union[str, Iterable[str]]] = None,
+        prefix: Optional[str] = None,
+        credential: _SDK_CREDENTIAL_TYPE = None,
+    ) -> AzureBlobStorageLoader:
```
kyleknap: For this factory function, I'm thinking we can simplify this by:
1. Remove the `account_url` and `container_name` parameters and just always use the defined fixture values for these. We can change this in the future if we have test cases that tweak these.
2. Just expose `prefix` and `credential` keyword arguments as `**kwargs` so that we can just proxy them to the initializer and not have to define these every time a new keyword argument gets added.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+    container_name: str,
+) -> Iterator[Tuple[MagicMock, MagicMock]]:
+    with patch(
+        "langchain_azure_storage.document_loaders.ContainerClient"
+    ) as mock_container_client_cls:
+        mock_client = MagicMock(spec=ContainerClient)
+        mock_client.list_blob_names.return_value = blob_names
+
+        def get_blob_client(blob_name: str) -> MagicMock:
+            mock_blob_client = MagicMock()
```
kyleknap: For any mocks we should make sure to include a spec. So here let's specify a `BlobClient`. Mainly it gives us a bit more type safety in a magic mock with no spec can swallow any missuses of interfaces.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+            mock_blob_client.url = f"{account_url}/{container_name}/{blob_name}"
+            mock_blob_data = MagicMock()
```
kyleknap: Same thing here. Let's make sure to add a spec to this magic mock.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+            blob_index = blob_names.index(blob_name)
```
kyleknap: Would it be possible to just turn this `get_blob_client()` internal helper function into a `mock_blob_clients` pytest fixture? I think my main concern is that in this approach we are starting to encode more logic into this mock (e.g., finding which blob from the list is being referred to) and may be difficult to maintain over time when we expect the blob order used to be constant.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+    container_name: str,
+    blob_names: list[str],
+    blob_contents: list[str],
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    loader = create_azure_blob_storage_loader()
+
+    for doc, expected_content, blob_name in zip(
+        loader.lazy_load(), blob_contents, blob_names
+    ):
```
kyleknap: Instead of `zip`, can we use `enumerate` for all of these cases? I think that will make it a little easier to follow as we will be tracking fewer variables needed for the assertion as we will just be able to index our `blobs` fixture for the expected blob names and blob contents.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+        assert doc.page_content == expected_content
+        assert doc.metadata["source"] == f"{account_url}/{container_name}/{blob_name}"
+
```
kyleknap, line 139: We should also assert that we fully iterated through the number of blobs that we expected to be loaded.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+    mock_client.get_blob_client.assert_called_once_with("text_file.txt")
+
+
+def test_default_credential(
+    mock_container_client: Tuple[MagicMock, MagicMock],
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    with patch.object(
+        azure.identity, "DefaultAzureCredential"
+    ) as mock_default_credential:
```
kyleknap: Instead of patching the `DefaultAzureCredential` can we just inspect what was provided to the patched container class to make sure it is an instance of the default credential object?

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def test_override_credential(
+    mock_container_client: Tuple[MagicMock, MagicMock],
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    from azure.core.credentials import AzureSasCredential
+
+    mock_container_client_cls, _ = mock_container_client
+    mock_credential = MagicMock(spec=AzureSasCredential)
```
kyleknap: Instead of using a MagicMock here, let's just provide the unmocked SAS credential as I'm not sure if we need any of the mocking capabilities.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+    loader = create_azure_blob_storage_loader(
+        blob_names="text_file.txt", credential=mock_credential
+    )
+
+    list(loader.lazy_load())
+    assert mock_container_client_cls.call_args[1]["credential"] == mock_credential
```
kyleknap: For this assertion, it probably makes sense to assert off its identity since this credential is expected to be directly used so asserting `is` as opposed to `==`

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+def test_async_credential_provided_to_sync(
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    from azure.core.credentials_async import AsyncTokenCredential
+
+    mock_credential = MagicMock(spec=AsyncTokenCredential)
+    with pytest.raises(TypeError):
+        loader = create_azure_blob_storage_loader(
+            blob_names="text_file.txt", credential=mock_credential
+        )
```
kyleknap: Let's move the loader instantiation outside of the pytest raises assertion to make it clear that we do not expect an exception to be raised at construction time here.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+    assert mock_container_client_cls.call_args[1]["credential"] == mock_credential
+
+
+def test_async_credential_provided_to_sync(
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    from azure.core.credentials_async import AsyncTokenCredential
+
+    mock_credential = MagicMock(spec=AsyncTokenCredential)
+    with pytest.raises(TypeError):
```
kyleknap: Let's also assert that raised exception includes a part of the message that we expect. Mainly this will ensure that a different `TypeError` than what we expected is being thrown since `TypeError` is a fairly general exception.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,182 @@
...
+
+def test_invalid_credential_type(
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    mock_credential = MagicMock(spec=str)
+    with pytest.raises(TypeError):
+        loader = create_azure_blob_storage_loader(
+            blob_names="text_file.txt", credential=mock_credential
+        )
+        list(loader.lazy_load())
```
kyleknap: Let's remove the `lazy_load()` here to make it clear that the exception raising is expected to come from the constructor.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        self._credential = self._get_sync_credential(self._credential)  # type: ignore
+        container_client = ContainerClient(
+            self._account_url, self._container_name, credential=self._credential
+        )
+        for blob_name in self._yield_blob_names(container_client):
+            blob_client = container_client.get_blob_client(blob_name)
+            yield from self._lazy_load_documents_from_blob(blob_client)
+
+    def _get_client_kwargs(self, credential: _SDK_CREDENTIAL_TYPE = None) -> dict:
```
kyleknap, line 72: This method does not appear to be used anymore. We should probably use it to set the `ContainerClient` keyword arguments because we'll likely need to reuse it for the async client when we get to that.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -0,0 +1,107 @@
...
+        else:
+            raise TypeError("Invalid credential type provided.")
+
+    def lazy_load(self) -> Iterator[Document]:
+        """Lazily load documents from Azure Blob Storage.
+
+        Yields:
+            the documents.
+        """
+        self._credential = self._get_sync_credential(self._credential)  # type: ignore
```
kyleknap: I'm thinking instead of mutating the `_credential` property, we do the following:
1. Rename `self._credential` to `self._provided_credential` to make it more clear that this represents what was provided
2. When we call `_get_sync_credential()` just store it as a normal variable (e.g., `credential` minus the `self`) and pass that into the `_get_client_kwargs` call.

I think that will help make it a little more readable and also ensure we are not mutating the instance of the class when invoking a load method, which could cause concurrency issues if used concurrently.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,162 @@
...
+        blob_names="text_file.txt", credential=mock_credential
+    )
+    with pytest.raises(ValueError, match="Cannot use synchronous load"):
+        list(loader.lazy_load())
+
+
+def test_invalid_credential_type(
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    mock_credential = MagicMock(spec=str)
```
kyleknap: Instead of wrapping this with a mock, let's just use a string (e.g., `"<account-key>") as we don't really need any mock functionality here

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,162 @@
...
+    list(loader.lazy_load())
+    assert mock_container_client_cls.call_args[1]["credential"] is mock_credential
+
+
+def test_async_credential_provided_to_sync(
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    from azure.core.credentials_async import AsyncTokenCredential
+
+    mock_credential = MagicMock(spec=AsyncTokenCredential)
```
kyleknap: Similar comment. Let's just use the async version of Azure Default credentials here instead of wrapping it with a mock since we really don't need the mock here.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,162 @@
...
+
+def test_lazy_load(
+    account_url: str,
+    container_name: str,
+    blobs: list[dict[str, str]],
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    loader = create_azure_blob_storage_loader()
+    docs = loader.lazy_load()
+    assert len(list(docs)) == len(blobs)
```
kyleknap: Instead of asserting the length before we iterate, it probably makes sense to make the assertion after the for loop and assert against the final value of the `index`? Mainly right now by calling the `list()` ahead of time, we exhaust the iterator and when we get to the for loop, none of the assertions are happening because there is nothing to iterate through.

Another idea we could do is maybe build up an expected document list and then assert the list returned from `lazy_load()` matches that expected list? This will only work if the `__eq__` is appropriately implemented on the `Document` base class. I think it is worth try out to see if it cleans some of this up.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+    return _create_azure_blob_storage_loader
+
+
+@pytest.fixture
+def mock_blob_clients(
+    account_url: str, container_name: str, blobs: list[dict[str, str]]
+) -> Callable[[str], MagicMock]:
+    def _get_blob_client(blob_name: str) -> MagicMock:
```
kyleknap, line 50: So instead of returning a callable, I was think we'd just return a list of mock blobs. The list should be able to be used downstream as a mock side effect too. If we want to keep this callable, we should rename the fixture `get_mock_blob_client` to make it a little more clear. I'm fine with either approach you prefer.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,162 @@
...
+        "langchain_azure_storage.document_loaders.ContainerClient"
+    ) as mock_container_client_cls:
+        mock_client = MagicMock(spec=ContainerClient)
+        mock_client.list_blob_names.return_value = [blob["blob_name"] for blob in blobs]
+        mock_client.get_blob_client.side_effect = mock_blob_clients
+        mock_container_client_cls.return_value = mock_client
+        yield mock_container_client_cls, mock_client
+
+
+def test_lazy_load(
```
kyleknap, line 94: We should also make sure to add two more test cases where a list of blobs is provided and a single blob as just a string. In these test cases, we should also assert that `list_blob_names` is not called.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -0,0 +1,162 @@
...
+            == f"{account_url}/{container_name}/{blobs[index]['blob_name']}"
+        )
+
+
+def test_get_blob_client(
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+    mock_container_client: Tuple[MagicMock, MagicMock],
+) -> None:
+    _, mock_client = mock_container_client
+    mock_client.list_blob_names.return_value = ["text_file.txt"]
```
kyleknap, line 146: Let's also assert after the `lazy_load()` is called that `list_blob_names` was called with the prefix of `text`.

kyleknap (commented):
Nice! This is looking great. Just had some smaller comments. I'm going to pull the implementation down next and play around with it, but otherwise, I think we are getting close to merging this.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -63,18 +63,17 @@ def lazy_load(self) -> Iterator[Document]:
...
         for blob_name in self._yield_blob_names(container_client):
             blob_client = container_client.get_blob_client(blob_name)
             yield from self._lazy_load_documents_from_blob(blob_client)
 
     def _get_client_kwargs(self, credential: _SDK_CREDENTIAL_TYPE = None) -> dict:
         return {
+            "account_url": self._account_url,
+            "container_name": self._container_name,
             "credential": credential,
-            "connection_data_block_size": self._CONNECTION_DATA_BLOCK_SIZE,
```
kyleknap: Let's keep this configuration value. It can speed up downloads significantly.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -145,7 +193,7 @@ def test_async_credential_provided_to_sync(
 def test_invalid_credential_type(
     create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
 ) -> None:
-    mock_credential = MagicMock(spec=str)
+    mock_credential = ""
```
kyleknap: Let's put in a value here like `<account-key>` to disambiguate the test case between testing whether the credential can be falsey or a string.

kyleknap (approved):
Looks good! :ship:

PR #156, opened 2025-09-30
Added document loaders loader_factory 

anjaliratnam-msft:
Implemented loader_factory for the document loader and added unit tests as well.

### DIFF
--- a/libs/azure-storage/langchain_azure_storage/document_loaders.py
+++ b/libs/azure-storage/langchain_azure_storage/document_loaders.py
@@ -104,16 +104,16 @@ class AzureBlobStorageLoader(BaseLoader):
     def _yield_documents_from_custom_loader(
         self, blob_content: bytes, blob_url: str
     ) -> Iterator[Document]:
-        with tempfile.NamedTemporaryFile(mode="wb", delete=False) as temp_file:
-            file_path = temp_file.name
-            temp_file.write(blob_content)
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_file_path = os.path.join(temp_dir, "temp_file.tmp")
+            with open(temp_file_path, "wb") as file:
+                file.write(blob_content)
 
-        if self._loader_factory is not None:
-            loader = self._loader_factory(temp_file.name)
-            for doc in loader.lazy_load():
-                doc.metadata["source"] = blob_url
-                yield doc
-        os.remove(file_path)
+            if self._loader_factory is not None:
+                loader = self._loader_factory(temp_file_path)
+                for doc in loader.lazy_load():
+                    doc.metadata["source"] = blob_url
+                    yield doc
 
     def _get_sync_credential(
         self, provided_credential: _SDK_CREDENTIAL_TYPE
--- a/libs/azure-storage/langchain_azure_storage/document_loaders.py
+++ b/libs/azure-storage/langchain_azure_storage/document_loaders.py
@@ -47,8 +47,10 @@ class AzureBlobStorageLoader(BaseLoader):
             credential: Credential to authenticate with the Azure Storage account.
                 If None, DefaultAzureCredential will be used.
             loader_factory: Optional callable that returns a custom document loader
-                (e.g. UnstructuredLoader) for parsing downloaded blobs. If None, content
-                will be returned as a single Document with UTF-8 text.
+                (e.g. UnstructuredLoader) for parsing downloaded blobs. If provided,
+                the blob contents will be downloaded to a temporary file whose name
+                gets passed to the callable. If None, content will be returned as a
+                single Document with UTF-8 text.
         """
         self._account_url = account_url
         self._container_name = container_name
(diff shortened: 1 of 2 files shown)


libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -42,7 +45,10 @@ def __init__(
             prefix: Prefix to filter blobs when listing from the container.
                 Cannot be used with blob_names.
             credential: Credential to authenticate with the Azure Storage account.
-                If None, DefaultAzureCredential will be used
+                If None, DefaultAzureCredential will be used.
+            loader_factory: Optional callable that returns a custom document loader
+                (e.g. UnstructuredLoader) for parsing downloaded blobs. If None, content
```
kyleknap: It might be good to add to the docstring that if a loader factory is provided it will download the blob contents to a temporary file and pass the temporary file name as a string to the callable. Otherwise, it is difficult to determine what the string parameter for the callable represents here.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -82,9 +90,22 @@ def _lazy_load_documents_from_blob(
...
+                file_path = temp_file.name
+
+            with open(file_path, "wb") as file:
+                file.write(blob_content)
+
+            loader = self._loader_factory(temp_file.name)
+            for doc in loader.lazy_load():
+                doc.metadata["source"] = blob_client.url
+                yield doc
+            os.remove(file_path)
```
kyleknap: Could we move all of this logic within the `NamedTemporaryFile()` context manager? Mainly the context manager will automatically handle deletion of the temporary file after it exits so we do not have to call `os.remove()` ourselves.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -82,9 +90,22 @@ def _lazy_load_documents_from_blob(
...
         blob_content = blob_data.readall()
-        yield Document(
-            blob_content.decode("utf-8"), metadata={"source": blob_client.url}
-        )
+        if self._loader_factory is None:
+            yield Document(
+                blob_content.decode("utf-8"), metadata={"source": blob_client.url}
+            )
+        else:
+            with tempfile.NamedTemporaryFile() as temp_file:
```
kyleknap: For readability, it might make sense to break the custom loader factory logic into its own separate method (e.g., `_yield_documents_from_custom_loader()`) and it can just accept the `blob_content`.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -20,12 +23,41 @@ def container_name() -> str:
     return "test-container"
 
 
+class CustomCSVLoader:
```
kyleknap: Let's subclass this from the `BaseLoader` class to make sure we are following the correct interface.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -20,12 +23,41 @@ def container_name() -> str:
...
+    def __init__(
```
kyleknap, line 29: It also might make sense to add a note as a comment that this custom loader follows the interface of the community csv loader: https://python.langchain.com/api_reference/community/document_loaders/langchain_community.document_loaders.csv_loader.CSVLoader.html and that we are not using it directly to avoid taking a dependency on the langchain community package.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -208,3 +256,49 @@ def test_both_blob_names_and_prefix_set(
...
+    container_name: str,
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+) -> None:
+    with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+        temp_file.write(b"col1,col2\nval1,val2\nval3,val4")
+
+    expected_csv_documents = get_expected_csv_documents(
+        temp_file.name, account_url, container_name
+    )
+
```
kyleknap: For both this test case and the one after, I'm thinking it would be worth just hardcoding what the expected Documents will be as a fixture (e.g., `expected_custom_csv_documents`) instead of trying to run the content through the loader to get the expected. Mainly suggesting this is as:
1. It shortens the test case bodies a bit
2. Makes it more explicit/readable on what the expected result from the custom CSV loader is
3. Guarantees that if we change the CSV loader in the future, we will have to acknowledge what the new transformation will be and possibly avoid a transformation of documents that matches the default loader document without us knowing.

libs/azure-storage/langchain_azure_storage/document_loaders.py
anjaliratnam-msft: I tried this but I get a permission denied error since it has to open the file when loader.lazy_load() is called.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -82,9 +92,28 @@ def _lazy_load_documents_from_blob(
...
+        else:
+            yield from self._yield_documents_from_custom_loader(
+                blob_content, blob_client.url
+            )
+
+    def _yield_documents_from_custom_loader(
+        self, blob_content: bytes, blob_url: str
+    ) -> Iterator[Document]:
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_file_path = os.path.join(temp_dir, "temp_file.tmp")
```
kyleknap: One subtle thing that I did not think of before, but we should probably have the temporary file name match what the blob name is. So for example if the blob name is `prefix/to/blob.txt` the file name used when creating the file should be `blob.txt`. We should be able to do this by using `os.path.basename` and the blob name from the blob client.

The reason that I'm suggesting this is that some loaders (especially ones that can support multiple file types) will use the extension to determine how to parse the file so by including the last filename component in the temporary file, it will help with that.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -82,9 +92,29 @@ def _lazy_load_documents_from_blob(
...
+        else:
+            yield from self._yield_documents_from_custom_loader(
+                blob_content, blob_client.url
+            )
+
+    def _yield_documents_from_custom_loader(
+        self, blob_content: bytes, blob_url: str
+    ) -> Iterator[Document]:
+        with tempfile.TemporaryDirectory() as temp_dir:
+            blob_name = os.path.basename(blob_url)
```
kyleknap: Can we use `blob_client.blob_name` instead? I'm hesitant to rely on the `blob_url` because there may be query strings attached to it and possibly url encoded. The blob name should better reflect what we want to be writing as the file name.

kyleknap (approved):
Looks great! :ship:

PR #167, opened 2025-10-06
GenAI tracing: compliant hierarchy, tool spans under chat, finish reason normalization, conversation id

nagkumar91:
This PR refactors AzureAI inference tracing to align with the GenAI span spec, reduce noise, and improve tool visibility. It introduces a single root invoke_agent per execution, parents
chats under that root, emits execute_tool spans under the chat that proposed them, normalizes finish reasons, and propagates a conversation id for easier grouping.

Key Changes

- Root hierarchy:
    - Creates a single top-level invoke_agent (fallback at chain_start), removes mid-execution “invoke_agent agent/tools” wrappers.
    - Parents chat spans directly under the root invoke_agent, even when parent context is missing or unknown.
- Tool spans:
    - Emits synthetic execute_tool spans under the chat when the assistant requests tools (message.tool_calls), including tool name, id, and parsed arguments.
    - Re-parents real on_tool_start spans under the most recent chat for the step; records tool_call.args and tool_call.result when content recording is enabled.
- Output/messages:
    - Normalizes gen_ai.response.finish_reasons: maps “tool_calls” → “tool_call”, defaults to “stop”.
    - Agent root keeps a role/parts transcript (opt-in) for step-level summary.
- Context:
    - Populates gen_ai.conversation.id with the root invoke_agent run_id and attaches it to chat and tool spans.
- Spec alignment:
    - Names chat spans “chat {model}” when model is available; sets server.address/port when present; provider override preserved (optional).

Tests Added/Updated

- Synthetic tool under chat: Verifies execute_tool spans parent under chat and include tool name; checks conversation id propagation.
- Suppress agent wrappers: Asserts on_agent_action doesn’t start invoke_agent.
- Tool span payload: Ensures tool span sets name/id/args/result and conversation id.
- Finish reason normalization: Asserts aggregated finish reasons normalize to singular.
- Chat parenting: Verifies chat spans parent under the root invoke_agent when no parent is supplied.

Validation

- Confirmed single trace per execution in App Insights with one root invoke_agent, chats under it, and execute_tool spans under chats.
- Verified normalized finish_reasons, token usage on chats, conversation id present on chats/tools.
- Tool span payloads (name/id/args/result) visible when content recording is enabled.

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
+++ b/libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
@@ -1188,7 +1188,6 @@ class AzureAIOpenTelemetryTracer(BaseCallbackHandler):
                 or _nm.startswith("tool")
             ):
                 return
-            return
         # Emit create_agent once per agent name (best-effort detection)
         try:
             if isinstance(name, str) and name and name not in self._created_agents:


libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
```diff
@@ -114,6 +114,16 @@ def _safe_json(obj: Any) -> str:
         return '"<unserializable>"'
 
 
+def _try_parse_json(s: Any) -> Any:
+    """Best-effort parse a JSON string, otherwise return the original value."""
+    if isinstance(s, str):
```
tonybaloney: Since `json.loads` takes str, bytes or bytearray, this should be more permissive to allow a byte string

```suggestion
    if isinstance(s, (str, bytes)):
```

nagkumar91: thanks!

PR #169, opened 2025-10-07
Document loaders integration tests

anjaliratnam-msft:
Added the integration tests for the document loader.

### DIFF
--- a/libs/azure-storage/tests/conftest.py
+++ b/libs/azure-storage/tests/conftest.py
@@ -20,6 +20,8 @@ def create_azure_blob_storage_loader(
     return _create_azure_blob_storage_loader
 
 
+# For the following expected csv document fixtures, the page content comes from
+# the tests.utils._TEST_BLOBS list.
 @pytest.fixture
 def expected_custom_csv_documents(
     account_url: str,
--- a/libs/azure-storage/tests/utils.py
+++ b/libs/azure-storage/tests/utils.py
@@ -4,6 +4,15 @@ from typing import Iterable, Iterator, Optional, Union
 from langchain_core.document_loaders import BaseLoader
 from langchain_core.documents.base import Document
 
+_TEST_BLOBS = [
+    {
+        "blob_name": "csv_file.csv",
+        "blob_content": "col1,col2\nval1,val2\nval3,val4",
+    },
+    {"blob_name": "json_file.json", "blob_content": "{'test': 'test content'}"},
+    {"blob_name": "text_file.txt", "blob_content": "test content"},
+]
+
 
 # This custom CSV loader follows the langchain_community.document_loaders.CSVLoader
 # interface. We are not directly using it to avoid adding langchain_community as a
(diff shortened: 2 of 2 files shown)


libs/azure-storage/tests/integration_tests/test_document_loaders.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@pytest.fixture(scope="session")
+def container_name() -> str:
+    return "document-loader-tests"
+
+
+@pytest.fixture(scope="session")
+def blob_service_client(account_url: str) -> BlobServiceClient:
+    return BlobServiceClient(account_url=account_url, credential=_CREDENTIAL)
```
kyleknap: We can probably declare the `DefaultAzureCredential()` here instead of at the module level since `_CREDENTIAL` is not used elsewhere.

libs/azure-storage/tests/integration_tests/test_document_loaders.py
```diff
@@ -0,0 +1,145 @@
...
+    blobs: list[dict[str, str]],
+    container_client: ContainerClient,
+) -> None:
+    for blob in blobs:
+        blob_client = container_client.get_blob_client(blob["blob_name"])
+        blob_client.upload_blob(blob["blob_content"], overwrite=True)
+
+
+@pytest.mark.parametrize(
+    "blob_names,expected_contents,prefix",
```
kyleknap: Instead of using `expected_contents` as the parameterization name, could we use `expected_blobs`? Mainly from the name of the variable, I'd expect `expected_contents` to be purely the contents of the blobs and not have additional metadata.

libs/azure-storage/tests/integration_tests/test_document_loaders.py
```diff
@@ -0,0 +1,145 @@
...
+                {
+                    "blob_name": "csv_file.csv",
+                    "blob_content": "col1,col2\nval1,val2\nval3,val4",
+                },
+                {
+                    "blob_name": "json_file.json",
+                    "blob_content": "{'test': 'test content'}",
+                },
+                {"blob_name": "text_file.txt", "blob_content": "test content"},
+            ],
```
kyleknap: I'm slightly concerned that if we update our `blobs` fixture to add a new blob that it will require us to update these test cases as well. I'm wondering if we could instead of having a `blobs` pytest fixture we just expose a utility called `get_test_blobs()` that will return all of the test blobs that we have and also allow us to filter which test blobs to return? This will allow us to use it in any parameterizations since it not a fixture and also help make these tests a bit less verbose too/more self descriptive.

libs/azure-storage/tests/utils.py
```diff
@@ -0,0 +1,79 @@
...
+    blob_names: Optional[Union[str, Iterable[str]]] = None, prefix: Optional[str] = None
+) -> list[dict[str, str]]:
+    blob_list = [
+        {
+            "blob_name": "csv_file.csv",
+            "blob_content": "col1,col2\nval1,val2\nval3,val4",
+        },
+        {"blob_name": "json_file.json", "blob_content": "{'test': 'test content'}"},
+        {"blob_name": "text_file.txt", "blob_content": "test content"},
+    ]
```
kyleknap: Let's maybe hoist this to the top of the `utils.py` file as a `_TEST_BLOBS` variable to make it really easy to find what the list of blobs are used in the test.

libs/azure-storage/tests/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return _create_azure_blob_storage_loader
+
+
+@pytest.fixture
+def expected_custom_csv_documents(
+    account_url: str,
+    container_name: str,
+) -> list[Document]:
+    return [
```
kyleknap, line 30: For both of these csv fixtures, it would help to add a small comment that the csv content is sourced from the `tests.utils._TEST_BLOBS` list to make it easier to locate where these expected document contents come from. Otherwise, I think with some of the refactoring we did, it is harder to track down where the expected contents of the documents came from since they are no longer in the same file anymore.

PR #170, opened 2025-10-09
Document loaders alazy_load()

anjaliratnam-msft:
Added the alazy_load() function for the document loader.

### DIFF
--- a/libs/azure-storage/tests/unit_tests/test_document_loaders.py
+++ b/libs/azure-storage/tests/unit_tests/test_document_loaders.py
@@ -99,7 +99,10 @@ def async_mock_container_client(
     ) as async_mock_container_client_cls:
 
         async def get_async_blob_names(**kwargs: Any) -> AsyncIterator[str]:
-            for blob_name in [blob["blob_name"] for blob in get_test_blobs()]:
+            prefix = kwargs.get("name_starts_with")
+            for blob_name in [
+                blob["blob_name"] for blob in get_test_blobs(prefix=prefix)
+            ]:
                 yield blob_name
 
         async_mock_client = AsyncMock(spec=AsyncContainerClient)
@@ -275,12 +278,6 @@ async def test_get_async_blob_client(
     async_mock_container_client: Tuple[AsyncMock, AsyncMock],
 ) -> None:
     _, async_mock_client = async_mock_container_client
-
-    async def updated_prefix_blob_list(**kwargs: Any) -> AsyncIterator[str]:
-        yield "text_file.txt"
-
-    async_mock_client.list_blob_names.side_effect = updated_prefix_blob_list
-
     loader = create_azure_blob_storage_loader(prefix="text")
     [doc async for doc in loader.alazy_load()]
     async_mock_client.get_blob_client.assert_called_once_with("text_file.txt")


libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -116,6 +150,38 @@ def _yield_documents_from_custom_loader(
...
+        self, async_blob_client: AsyncBlobClient
+    ) -> AsyncIterator[Document]:
+        blob_data = await async_blob_client.download_blob(
+            max_concurrency=self._MAX_CONCURRENCY
+        )
+        blob_content = await blob_data.readall()
+        if self._loader_factory is None:
+            yield Document(
+                blob_content.decode("utf-8"), metadata={"source": async_blob_client.url}
+            )
```
kyleknap: It might make sense to make this a small helper function (e.g. `_get_default_document()`) that accepts both the blob content and the client so that we do not have to reproduce this in multiple methods.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -132,8 +198,36 @@ def _get_sync_credential(
...
+    ) -> _SDK_CREDENTIAL_TYPE:
+        if provided_credential is None:
+            return azure.identity.aio.DefaultAzureCredential()
+        if not isinstance(
+            provided_credential, azure.core.credentials_async.AsyncTokenCredential
+        ):
+            raise ValueError(
+                "Cannot use asynchronous load methods when AzureBlobStorageLoader is "
+                "instantiated using a synchronous TokenCredential or "
+                "AzureSasCredential. Use its synchronous load method instead or supply"
```
kyleknap: Technically, we should be able to support `AzureSasCredential` since there is no I/O involved with the class. Instead, we should probably be detecting if a synchronous token credential was provided as the class already validates as that the credential is broadly supported in the loader.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -79,6 +91,28 @@ def lazy_load(self) -> Iterator[Document]:
...
+            async for blob_name in self._ayield_blob_names(async_container_client):
+                async_blob_client = async_container_client.get_blob_client(blob_name)
+                async for doc in self._alazy_load_documents_from_blob(
+                    async_blob_client
+                ):
+                    yield doc
+            if isinstance(
+                credential, azure.core.credentials_async.AsyncTokenCredential
+            ):
+                await credential.close()
```
kyleknap: So, something really subtle but we only want to close the credential if the loader was the one that created it. So the pattern should be is:
1. If the loader created the credential, we want to make sure to close it. This is because the customer won't be able to close it on their own.
2. If the customer provided a credential, we don't want to close it and it is their responsibility to close it. This is important because the credential could be shared across clients so if the loader closes the credential, it could cause other clients to stop working.

In terms of structuring this, we should:
1. Probably check whether the `credential is self._provided_credential` to determine if the loader created it or not.
2. Split the logic into two if branches where in one we use a context manager on the credential that the loader created and one when no context manager is used. This will probably look not great but it is important to use the context manager so that if there is an error in the logic, we still close the credential that we created.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -116,6 +150,38 @@ def _yield_documents_from_custom_loader(
...
+                yield doc
+
+    async def _ayield_documents_from_custom_loader(
+        self, blob_content: bytes, async_blob_client: AsyncBlobClient
+    ) -> AsyncIterator[Document]:
+        with tempfile.TemporaryDirectory() as temp_dir:
+            blob_name = os.path.basename(async_blob_client.blob_name)
+            temp_file_path = os.path.join(temp_dir, blob_name)
+            with open(temp_file_path, "wb") as file:
+                file.write(blob_content)
```
kyleknap: This might get a little messy but I think it is something that we'll need to address and had not thought about before... Right now this all of this I/O is synchronous meaning that when called it will block the event loop which works against being able to do many tasks concurrently. Furthermore, the standard library does not provide async interfaces for file operations because native file operations are synchronous.

That being said, I suspect the write to disk call here could take a significant amount of time and something we likely don't want to block the event loop on. The way to get around this is to have this logic ran in a thread to unblock the event loop. This may be tricky, but I think it would be worth exploring the following in this order:
1. Use LangChain's [`run_in_executor()`](https://python.langchain.com/api_reference/core/runnables/langchain_core.runnables.config.run_in_executor.html) utility. It is used frequently for adding async methods to base classes. So, I'd recommend looking through the langchain core code for examples.
2. Take a dependency on [`aiofiles`](https://github.com/Tinche/aiofiles). It implements the wrapping logic that we'd need but I'd like us to avoid taking on a 3p dependency as best we can.

libs/azure-storage/tests/utils.py
```diff
@@ -42,6 +42,17 @@ def lazy_load(self) -> Iterator[Document]:
                     page_content=content, metadata={"source": self.file_path}
                 )
 
+    async def alazy_load(self) -> AsyncIterator[Document]:
+        for doc in self.lazy_load():
+            yield doc
```
kyleknap: Given the base loader class already implements a default `alazy_load()`, I think we can reuse it especially since at best we'd throw this into an executor as well.

libs/azure-storage/tests/utils.py
```diff
@@ -42,6 +42,17 @@ def lazy_load(self) -> Iterator[Document]:
...
+
+
+def csv_loader_factory(file_path: str) -> CustomCSVLoader:
```
kyleknap: Could rename this to something more self descriptive like `get_first_column_csv_loader`? Mainly now that it is a utility, it is a little more difficult to understand what the expected behavior of the loader in the tests without looking at `utils.py`.

libs/azure-storage/tests/utils.py
```diff
@@ -43,6 +43,13 @@ def lazy_load(self) -> Iterator[Document]:
                 )
 
 
+def get_first_column_csv_loader_factory(file_path: str) -> CustomCSVLoader:
```
kyleknap: Really subtle on the naming, but as is, the function name seems to indicate that it returns a factory, but instead in reality we want the function to **be** the factory. Instead, could we rename it `get_first_column_csv_loader()` or `first_column_csv_loader_factory()`. These names better indicate the function itself is the factory and it will create our specialized loader when called.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -98,20 +100,37 @@ async def alazy_load(self) -> AsyncIterator[Document]:
...
-                ):
-                    yield doc
-            if isinstance(
-                credential, azure.core.credentials_async.AsyncTokenCredential
-            ):
-                await credential.close()
+        if self._provided_credential is None and isinstance(
+            credential, azure.core.credentials_async.AsyncTokenCredential
+        ):
+            async with credential:
```
kyleknap: Sorry for the churn but had an idea after looking at this a second time... Instead of having the if branching directly in the `alazy_load()`, I'm wondering if we can leverage an approach similar to the `asyncontextmanager` that we introduced where we can just make `_get_async_credential` an async context manager? Specifically, if the loader creates the default credentials, it will yield within its own context manager and if it is a provided credential, we just yield it back directly without wrapping it with a `with` context manager statement? I think this could really help clean up the logic here.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -204,13 +235,17 @@ def _get_async_credential(
         if provided_credential is None:
             return azure.identity.aio.DefaultAzureCredential()
         if not isinstance(
-            provided_credential, azure.core.credentials_async.AsyncTokenCredential
+            provided_credential,
+            (
+                azure.core.credentials_async.AsyncTokenCredential,
```
kyleknap, line 239: Can we just check whether the credential is a sync token credential instead? e.g.,
```python
if isinstance(provided_credential, azure.core.credential.TokenCredential)
```
It should help make this logic a bit simpler in case we add support for credentials in the future that do not differentiate between sync and async.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -170,18 +185,34 @@ async def _alazy_load_documents_from_blob(
...
                     yield doc
 
+    @asynccontextmanager
+    async def _temp_file_context_manager(
+        self, blob_content: bytes, async_blob_client: AsyncBlobClient
+    ) -> AsyncIterator[str]:
+        temp_dir = tempfile.TemporaryDirectory()
+        try:
+            blob_name = os.path.basename(async_blob_client.blob_name)
+            temp_file_path = os.path.join(temp_dir.name, blob_name)
```
kyleknap: I'm wonder if:
1. We can move these into the `_write_to_tempfile()` method
2. Reuse this method in the sync implementation now that we have broken into a new method

That should help us reuse the logic a bit better across the sync and async implementations.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -170,18 +185,34 @@ async def _alazy_load_documents_from_blob(
...
                 loader = self._loader_factory(temp_file_path)
                 async for doc in loader.alazy_load():
                     doc.metadata["source"] = async_blob_client.url
                     yield doc
 
+    @asynccontextmanager
+    async def _temp_file_context_manager(
+        self, blob_content: bytes, async_blob_client: AsyncBlobClient
+    ) -> AsyncIterator[str]:
+        temp_dir = tempfile.TemporaryDirectory()
```
kyleknap: When creating the `TemporaryDirectory` and cleaning it up, we should probably run these two in an executor as well

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -170,18 +185,34 @@ async def _alazy_load_documents_from_blob(
...
+            blob_content, async_blob_client
+        ) as temp_file_path:
             if self._loader_factory is not None:
                 loader = self._loader_factory(temp_file_path)
                 async for doc in loader.alazy_load():
                     doc.metadata["source"] = async_blob_client.url
                     yield doc
 
+    @asynccontextmanager
+    async def _temp_file_context_manager(
```
kyleknap: Maybe let's rename this to `_blob_content_as_temp_file`? Mainly it is a bit more self-descriptive on what the temp file represents when looking at it from methods that use it.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -170,18 +185,34 @@ async def _alazy_load_documents_from_blob(
...
+        temp_dir = tempfile.TemporaryDirectory()
+        try:
+            blob_name = os.path.basename(async_blob_client.blob_name)
+            temp_file_path = os.path.join(temp_dir.name, blob_name)
+            await run_in_executor(
+                None, self._write_to_temp_file, blob_content, temp_file_path
+            )
+            yield temp_file_path
+        finally:
+            temp_dir.__exit__(None, None, None)
```
kyleknap: Instead of calling `__exit__()`, let's just call its [`cleanup()`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory.cleanup) method. Mainly, the `cleanup()` method is intended to be the interface to use if you want to cleanup the directory manually and looks cleaner. If this method did not exist, we'd then probably have to resort to manually calling `__exit__()`

libs/azure-storage/langchain_azure_storage/document_loaders.py
anjaliratnam-msft: I tried that but then the ValueError gets raised when the async credential is passed in. I could add another if to check if it's AsyncTokenCredential and return it before checking if it's sync if that makes more sense?

kyleknap: Oh interesting... Let's keep the logic as is then and maybe add a comment about that behavior for reference.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -178,13 +229,119 @@ def test_custom_loader_factory_with_configurations(
...
+async def test_get_async_blob_client(
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+    async_mock_container_client: Tuple[AsyncMock, AsyncMock],
+) -> None:
+    _, async_mock_client = async_mock_container_client
+
+    async def updated_prefix_blob_list(**kwargs: Any) -> AsyncIterator[str]:
+        yield "text_file.txt"
+
+    async_mock_client.list_blob_names.side_effect = updated_prefix_blob_list
```
kyleknap: In our `async_mock_container_client`, could we just proxy the `prefix` kwarg into `get_test_blobs()` to avoid needing to define this logic?

PR #176, opened 2025-10-14
Added user agent

anjaliratnam-msft:
Added the user agent for langchain-azure-storage.

### DIFF
--- a/libs/azure-storage/tests/unit_tests/test_document_loaders.py
+++ b/libs/azure-storage/tests/unit_tests/test_document_loaders.py
@@ -353,5 +353,17 @@ def test_user_agent(
     user_agent = f"azpartner-langchain/{__version__}"
     loader = create_azure_blob_storage_loader(blob_names="text_file.txt")
     list(loader.lazy_load())
-    headers = mock_container_client_cls.call_args[1]
-    assert headers["user_agent"] == user_agent
+    client_kwargs = mock_container_client_cls.call_args[1]
+    assert client_kwargs["user_agent"] == user_agent
+
+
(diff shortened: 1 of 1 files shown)


libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -342,3 +343,15 @@ async def test_async_custom_loader_factory_with_configurations(
...
+def test_user_agent(
+    create_azure_blob_storage_loader: Callable[..., AzureBlobStorageLoader],
+    mock_container_client: Tuple[MagicMock, MagicMock],
+) -> None:
+    mock_container_client_cls, _ = mock_container_client
+    user_agent = f"azpartner-langchain/{__version__}"
+    loader = create_azure_blob_storage_loader(blob_names="text_file.txt")
+    list(loader.lazy_load())
+    headers = mock_container_client_cls.call_args[1]
+    assert headers["user_agent"] == user_agent
```
kyleknap: Small nitpick, but let's name the variable name here `client_kwargs` as the value passed hear is not directly the user agent value and is more a client configuration value.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -342,3 +343,15 @@ async def test_async_custom_loader_factory_with_configurations(
     assert [
         doc async for doc in loader.alazy_load()
     ] == expected_custom_csv_documents_with_columns
+
+
+def test_user_agent(
```
kyleknap, line 348: Can we add a test case for the async client as well?

PR #178, opened 2025-10-16
Remove duplicate tool call span

nagkumar91:


marlenezw:
Nice! No issues from me! Going to merge

PR #180, opened 2025-10-16
Document loaders datalake support

anjaliratnam-msft:
Added support to now exclude directories from being returned as documents when a DataLake account is used.

### DIFF
--- a/libs/azure-storage/tests/integration_tests/test_document_loaders.py
+++ b/libs/azure-storage/tests/integration_tests/test_document_loaders.py
@@ -184,11 +184,9 @@ class TestDataLakeDirectoryFiltering:
             "document-loader-tests"
         )
         container_client.create_container()
-        for blob in get_datalake_test_blobs(include_directories=True):
+        for blob in get_datalake_test_blobs():
             blob_client = container_client.get_blob_client(blob["blob_name"])
-            blob_client.upload_blob(
-                blob["blob_content"], metadata=blob["metadata"], overwrite=True
-            )
+            blob_client.upload_blob(blob["blob_content"], overwrite=True)
 
         yield
         container_client.delete_container()
--- a/libs/azure-storage/tests/unit_tests/test_document_loaders.py
+++ b/libs/azure-storage/tests/unit_tests/test_document_loaders.py
@@ -138,7 +138,7 @@ def async_mock_container_client(
 
         async def get_async_blobs(**kwargs: Any) -> AsyncIterator[MagicMock]:
             prefix = kwargs.get("name_starts_with")
-            for mock_blob in get_test_mock_blobs(get_test_blobs(), prefix):
+            for mock_blob in get_test_mock_blobs(get_test_blobs(prefix=prefix)):
                 yield mock_blob
 
         async_mock_client = AsyncMock(spec=AsyncContainerClient)
@@ -180,7 +180,7 @@ def async_mock_datalake_container_client(
         async def get_async_blobs(**kwargs: Any) -> AsyncIterator[MagicMock]:
             prefix = kwargs.get("name_starts_with")
             for mock_blob in get_test_mock_blobs(
-                get_datalake_test_blobs(include_directories=True), prefix=prefix
+                get_datalake_test_blobs(prefix=prefix, include_directories=True)
             ):
                 yield mock_blob
 
--- a/libs/azure-storage/tests/utils.py
+++ b/libs/azure-storage/tests/utils.py
@@ -96,32 +96,16 @@ def get_datalake_test_blobs(
 ) -> list[dict[str, Any]]:
     if not include_directories:
         return get_test_blobs(blob_names, prefix)
-
-    blobs = _get_test_blobs(_TEST_DATALAKE_BLOBS, blob_names, prefix)
-    return [
-        {
-            **blob,
-            "size": len(blob["blob_content"]),
-            "metadata": blob.get("metadata", None),
-        }
-        for blob in blobs
-    ]
+    return _get_test_blobs(_TEST_DATALAKE_BLOBS, blob_names, prefix)
 
 
-def get_test_mock_blobs(
-    blob_list: list[dict[str, Any]], prefix: Optional[str] = None
-) -> list[MagicMock]:
-    updated_blobs = [
-        blob
-        for blob in blob_list
-        if prefix is None or blob["blob_name"].startswith(prefix)
-    ]
+def get_test_mock_blobs(blob_list: list[dict[str, Any]]) -> list[MagicMock]:
     mock_blobs = []
-    for blob in updated_blobs:
+    for blob in blob_list:
         mock_blob = MagicMock(spec=BlobProperties)
         mock_blob.name = blob["blob_name"]
         mock_blob.size = len(blob["blob_content"])
-        mock_blob.metadata = blob.get("metadata", {})
+        mock_blob.metadata = blob.get("metadata", None)
         mock_blobs.append(mock_blob)
 
     return mock_blobs
@@ -139,11 +123,26 @@ def _get_test_blobs(
         for name in blob_names:
             for blob in blob_list:
                 if blob["blob_name"] == name:
-                    updated_list.append(blob)
+                    updated_list.append(
+                        {
+                            **blob,
+                            "size": len(blob["blob_content"]),
+                            "metadata": blob.get("metadata", None),
+                        }
+                    )
                     break
         return updated_list
 
-    if prefix is not None:
-        return [blob for blob in blob_list if blob["blob_name"].startswith(prefix)]
-
-    return blob_list
+    prefix_blob_list = (
+        [blob for blob in blob_list if blob["blob_name"].startswith(prefix)]
+        if prefix
+        else blob_list
+    )
+    return [
+        {
+            **blob,
+            "size": len(blob["blob_content"]),
+            "metadata": blob.get("metadata", None),
+        }
+        for blob in prefix_blob_list
+    ]


libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -253,7 +253,9 @@ def _yield_blob_names(self, container_client: ContainerClient) -> Iterator[str]:
         if self._blob_names is not None:
             yield from self._blob_names
         else:
-            yield from container_client.list_blob_names(name_starts_with=self._prefix)
+            for blob in container_client.list_blobs(name_starts_with=self._prefix):
```
kyleknap: Ok so one more thing, we'll want to do is:
1. Add `include="metadata"` to that `list_blobs` call. This will ensure we get blob metadata coming back
2. Check for the `hdi_isfolder` metadata key and that it equals the string `"true"` to indicate that this is an adls folder

We should be able to rely on this. I think there may be a more explicit response header value to check, `x-ms-resource-type`, but it is not exposed in the SDK yet. We can move over to that if it gets exposed in the future.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -253,7 +253,9 @@ def _yield_blob_names(self, container_client: ContainerClient) -> Iterator[str]:
...
+                if blob.size > 0:
```
kyleknap: When we go to add the additional `hdi_isfolder` check, let's make sure to break this down into a helper method `_is_adls_directory()` to help with readability.

libs/azure-storage/tests/utils.py
```diff
@@ -8,9 +8,14 @@
...
         "blob_name": "csv_file.csv",
         "blob_content": "col1,col2\nval1,val2\nval3,val4",
+        "size": "32",
     },
-    {"blob_name": "json_file.json", "blob_content": "{'test': 'test content'}"},
-    {"blob_name": "text_file.txt", "blob_content": "test content"},
+    {
+        "blob_name": "json_file.json",
+        "blob_content": "{'test': 'test content'}",
+        "size": "24",
```
kyleknap: Instead of hardcoding the size here, could we programatically add it as part of the `get_test_blobs()`? It might be easier to add more blobs in the future. Also, we should make sure these sizes are represented as integers and not strings.

libs/azure-storage/tests/integration_tests/test_document_loaders.py
```diff
@@ -57,6 +57,42 @@ def upload_blobs_to_container(
...
+
+@pytest.fixture
+def datalake_container_setup(
+    datalake_blob_service_client: BlobServiceClient,
+) -> Iterator[None]:
+    container_client = datalake_blob_service_client.get_container_client(
+        "document-loader-tests"
+    )
+    container_client.create_container()
+    for blob in get_test_blobs():
```
kyleknap: Instead of directly reusing the `get_test_blobs()`, it may make sense to expose a `get_test_datalake_blobs()` and `_TEST_DATALAKE_BLOBS` where we just reuse the blobs from `_TEST_BLOBS` and just add on a single blob that is in a directory. This will help keeping all of the datalake added logic consolidated to a single spot and out of the test case bodies.

libs/azure-storage/tests/integration_tests/test_document_loaders.py
```diff
@@ -149,3 +185,23 @@ async def test_alazy_load_with_loader_factory_configurations(
     assert [
         doc async for doc in loader.alazy_load()
     ] == expected_custom_csv_documents_with_columns
+
+
+def test_datalake_excludes_directories(
```
kyleknap: We should make sure to have an async version of this test as well since there are code paths in the async code path that is separate from the sync code path.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -61,9 +61,14 @@ def mock_container_client(
...
-            blob["blob_name"] for blob in get_test_blobs()
-        ]
+        mock_blobs = []
+        for blob in get_test_blobs():
+            mock_blob = MagicMock()
+            mock_blob.name = blob["blob_name"]
+            mock_blob.size = int(blob["size"])
+            mock_blobs.append(mock_blob)
+
+        mock_client.list_blobs.return_value = mock_blobs
```
kyleknap: Would it be possible to add a unit test for both the async and sync version where we ensure that we do not return blobs that match an adls directory? While we have the integration tests, a unit test version is useful as well to capture what the exact heuristic is and also ensure it is always ran in CI since the integration tests are not ran in CI as of now and not as ran as often.

kyleknap (commented):
Looks good! Just had some smaller comments, but overall is looking solid.

libs/azure-storage/langchain_azure_storage/document_loaders.py
```diff
@@ -265,14 +269,18 @@ async def _ayield_blob_names(
...
 
     def _get_default_document(
         self, blob_content: bytes, blob_client: Union[BlobClient, AsyncBlobClient]
     ) -> Document:
         return Document(
             blob_content.decode("utf-8"), metadata={"source": blob_client.url}
         )
+
+    def _is_adls_directory(self, blob: BlobProperties) -> bool:
+        return blob.size == 0 and blob.metadata.get("hdi_isfolder") == "true"
```
kyleknap: In between the `blob.size` and the `hdi_isfolder` check, let's add a `blob.metadata`. So:
```python
blob.size == 0 and blob.metadata and blob.metadata.get("hdi_isfolder") == "true"
```
Mainly blobs could be empty and not be directories and have `metadata` that equates to `None`. This additional clause will make sure we short-circuit early to avoid trying to call `.get()` on a `None` value and cause the entire listing to error out.

libs/azure-storage/tests/utils.py
```diff
@@ -84,3 +99,52 @@ def get_test_blobs(
         return [blob for blob in _TEST_BLOBS if blob["blob_name"].startswith(prefix)]
 
     return _TEST_BLOBS
+
+
+def get_datalake_test_blobs(
+    blob_names: Optional[Union[str, Iterable[str]]] = None, prefix: Optional[str] = None
+) -> list[dict[str, Any]]:
+    if blob_names is not None:
```
kyleknap, line 119: Instead of duplicating the code for `get_test_blobs()`, I'm wondering if we can expose a new `_get_test_blobs()` utility that both `get_test_blobs()` and `get_datalake_test_blobs()` can call out to by passing in the list of blobs to use? Mainly I think this can help reused code especially if we want to expand on the scaffolding in the future.

libs/azure-storage/tests/utils.py
```diff
@@ -9,8 +9,23 @@
...
         "blob_content": "col1,col2\nval1,val2\nval3,val4",
     },
-    {"blob_name": "json_file.json", "blob_content": "{'test': 'test content'}"},
-    {"blob_name": "text_file.txt", "blob_content": "test content"},
+    {
+        "blob_name": "json_file.json",
+        "blob_content": "{'test': 'test content'}",
+    },
+    {
+        "blob_name": "text_file.txt",
```
kyleknap: For completeness, maybe let's make this blob name `directory/test_file.txt` so it fits the more common pattern when we have there is an ADLS directory and blob underneath of it when we get to the datalake blobs list?

libs/azure-storage/tests/utils.py
```diff
@@ -84,3 +99,52 @@ def get_test_blobs(
...
+            blob_names = [blob_names]
+        updated_list = []
+        for name in blob_names:
+            for blob in _TEST_DATALAKE_BLOBS:
+                if blob["blob_name"] == name:
+                    updated_list.append(
+                        {
+                            **blob,
+                            "size": len(blob["blob_content"]),
+                            "metadata": blob.get("metadata", {}),
```
kyleknap: For any of the metadata, if it is missing let's set the value to `None` instead of empty dictionary as it seems like when testing against the blob service the end value from the SDK is `None` if there is no metadata.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -99,20 +148,67 @@ def async_mock_container_client(
...
+    ) as async_mock_container_client_cls:
+
+        async def get_async_blobs(**kwargs: Any) -> AsyncIterator[MagicMock]:
+            prefix = kwargs.get("name_starts_with")
+            for blob in get_datalake_test_blobs(prefix=prefix):
+                mock_blob = MagicMock()
+                mock_blob.name = blob["blob_name"]
+                mock_blob.size = int(blob["size"])
+                mock_blob.metadata = blob["metadata"]
+                yield mock_blob
```
kyleknap: I'm wondering if it makes sense if we take the `BlobProperty` scaffolding logic and roll it up into the `get_test_blobs()` and `get_datalake_test_blobs()` utilities.  And in order to get mocks back we could either expose an `as_mock` boolean or expose it as new utilities `get_test_mock_blobs()`. I don't have a strong preference on either option, but I think that should be able to help us consolidate some of the logic in all of the places we need to build up these mock blobs, which has expanded quite a bit now.

libs/azure-storage/tests/unit_tests/test_document_loaders.py
```diff
@@ -99,20 +148,67 @@ def async_mock_container_client(
...
 
-        async def get_async_blob_names(**kwargs: Any) -> AsyncIterator[str]:
+        async def get_async_blobs(**kwargs: Any) -> AsyncIterator[MagicMock]:
             prefix = kwargs.get("name_starts_with")
-            for blob_name in [
-                blob["blob_name"] for blob in get_test_blobs(prefix=prefix)
-            ]:
-                yield blob_name
+            for blob in get_test_blobs(prefix=prefix):
+                mock_blob = MagicMock()
```
kyleknap: When mocking the blob properties, we should use be throwing in a `BlobProperties` spec to give a little more safety on public interfaces.

libs/azure-storage/tests/utils.py
```diff
@@ -84,3 +99,52 @@ def get_test_blobs(
...
+        {
+            **blob,
+            "size": len(blob["blob_content"]),
+            "metadata": blob.get("metadata", {}),
+        }
+        for blob in _TEST_DATALAKE_BLOBS
+    ]
+
+
+def get_expected_datalake_blobs() -> list[dict[str, Any]]:
```
kyleknap: Instead of exposing a new utility, let's add a new parameter `include_directories` boolean that defaults to `False` to the `get_datalake_test_blobs()` to make it a little more explicit when ADLS directories are included in the list.

libs/azure-storage/tests/integration_tests/test_document_loaders.py
```diff
@@ -185,7 +184,7 @@ def datalake_container_setup(
             "document-loader-tests"
         )
         container_client.create_container()
-        for blob in get_datalake_test_blobs():
+        for blob in get_datalake_test_blobs(include_directories=True):
```
kyleknap: For the datalake tests, should we be including the directories in what we upload? To my understanding, ADLS will automatically create the directory and that is more representative of how customers will have ADLS directories in the first place.

libs/azure-storage/tests/utils.py
```diff
@@ -84,67 +86,64 @@ def get_expected_documents(
...
+        blob
+        for blob in blob_list
+        if prefix is None or blob["blob_name"].startswith(prefix)
+    ]
+    mock_blobs = []
+    for blob in updated_blobs:
+        mock_blob = MagicMock(spec=BlobProperties)
+        mock_blob.name = blob["blob_name"]
+        mock_blob.size = len(blob["blob_content"])
+        mock_blob.metadata = blob.get("metadata", {})
```
kyleknap: Let's set the metadata value to `None` if it is not present e.g.:
```python
mock_blob.metadata = blob.get("metadata", None)
```

libs/azure-storage/tests/utils.py
```diff
@@ -69,18 +86,64 @@ def get_expected_documents(
...
+
+    blobs = _get_test_blobs(_TEST_DATALAKE_BLOBS, blob_names, prefix)
+    return [
+        {
+            **blob,
+            "size": len(blob["blob_content"]),
+            "metadata": blob.get("metadata", None),
+        }
+        for blob in blobs
+    ]
```
kyleknap: For the `size` and `metadata`, we should just lower this into our `_get_test_blobs()` helper function so that we are consistent on what data is returned. In general, size and metadata will be returned no matter the account type.

libs/azure-storage/tests/utils.py
```diff
@@ -69,18 +86,64 @@ def get_expected_documents(
...
+
+
+def get_test_mock_blobs(
+    blob_list: list[dict[str, Any]], prefix: Optional[str] = None
+) -> list[MagicMock]:
+    updated_blobs = [
+        blob
+        for blob in blob_list
+        if prefix is None or blob["blob_name"].startswith(prefix)
+    ]
```
kyleknap: Do we need to expose `prefix` logic in this helper method? I'd assume because `prefix` is exposed in both the `get_test_blobs` and `get_test_datalake_blobs` utilities that we would not need it.

kyleknap (commented):
Looks good. Just a little more feedback and we should be set.

kyleknap (approved):
Looks good! :ship:

PR #182, opened 2025-10-16
Add azure storage support for LangChain 1.0

kyleknap:
Changes include:
* Updating package version to 1.0 to mirror LangChain core bump to 1.0
* Remove support for Python 3.9 since LangChain 1.0 drops support
* Update dependencies to rely on langchain-core 1.0 release. Also, bulk updated dependency locks to make sure we are running CI on latest version of dependencies.
* Update package classifier
* Add beta decorator to document loader. This will allow us to get feedback on the interface before marking it as stable.

The beta warning message looks like this when the loader is instantiated:
```
C:\Users\kyleknapp\GitHub\langchain-azure\libs\azure-storage\demos\app.py:14: LangChainBetaWarning: `AzureBlobStorageLoader` is in public preview. Its API is not stable and may change in future versions.
  loader = AzureBlobStorageLoader(
```

### DIFF
--- a/libs/azure-storage/pyproject.toml
+++ b/libs/azure-storage/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "langchain-azure-storage"
-version = "1.0.0rc1"
+version = "1.0.0"
 description = "LangChain integrations for Azure Storage"
 authors = [
     {name="Microsoft Corporation", email="ascl@microsoft.com"},
@@ -24,7 +24,7 @@ classifiers = [
 dependencies = [
     "azure-identity~=1.25",
     "azure-storage-blob[aio]~=12.26",
-    "langchain-core>=1.0.0rc2,<2.0.0",
+    "langchain-core~=1.0",
 ]
 
 [project.urls]
--- a/libs/azure-storage/poetry.lock
+++ b/libs/azure-storage/poetry.lock
@@ -14,132 +14,132 @@ files = [
 
 [[package]]
 name = "aiohttp"
-version = "3.13.0"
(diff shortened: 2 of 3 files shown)


libs/azure-storage/pyproject.toml
```diff
@@ -1,21 +1,20 @@
...
     {name="Microsoft Corporation", email="ascl@microsoft.com"},
 ]
 license = { file = "LICENSE" }
 readme = "README.md"
-requires-python = "~=3.9"
+requires-python = "~=3.10"
 keywords = ["langchain", "azure"]
 classifiers = [
-    "Development Status :: 1 - Planning",
+    "Development Status :: 4 - Beta",
```
kyleknap, line 13: @marlenezw one thing I was on the fence about was whether to use the Beta PyPI classifier here or the `Development Status :: 5 - Production/Stable` classifier? Typically, if the package is a 1.0, I'd expect it to have the production classifier. However, given this package will only have one feature that is decorated as a `beta`, it felt awkward to use the production classifier. I'm fine switching it though if the expectation that all of the langchain azure packages should be marked as production/stable to mirror LangChain core. This decision does not functionally impact the package; its more comes down to how we are looking to label the package on PyPI. Would love to get your opinion here.

anjaliratnam-msft (approved):

libs/azure-storage/pyproject.toml
marlenezw: Lets use beta! Great catch and I'll add this to the wider chat so some folks can use beta too if they are on the fence! Thank you @kyleknap

PR #197, opened 2025-10-27
Tracer semconv alignment for multi agent

nagkumar91:
Updating the tracer to create a parent trace for a multi agent scenrio.
Previously, the sample running multi agent was required to create a parent. 
Enables traces like this:
Complex trace screnshot

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
+++ b/libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
@@ -688,6 +688,7 @@ class AzureAIOpenTelemetryTracer(BaseCallbackHandler):
                     parent_record.attributes.get(Attrs.USAGE_TOTAL_TOKENS)
                 )
 
+                updated_input: Optional[int] = existing_input
                 delta_input: Optional[int] = None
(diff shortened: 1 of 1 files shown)


libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
```diff
@@ -560,8 +594,17 @@ def __init__(
...
     ) -> bool:
+        if metadata and metadata.get("otel_agent_span"):
+            node_name = metadata.get("langgraph_node")
+            meta_agent_name = metadata.get("agent_name") or metadata.get("agent_type")
+            if node_name and meta_agent_name and str(node_name) != str(meta_agent_name):
+                return True
+            return False
         if agent_name == "LangGraph":
             return False
         if parent_run_id is None:
```
santiagxf, line 614: @nagkumar91 shouldn't this return True? I see in the trace that `LangGraph` is identified as an agent.

nagkumar91: Thats the default name, for the root span, the callback is not returning an `agent_name`

PR #204, opened 2025-11-10
Fix inconsistent text field aliasing in vector and hybrid search queries

davide-evangelisti-prom:
## Summary

This PR fixes bugs in the `AzureCosmosDBNoSqlVectorSearch` class where the text field alias is hardcoded as `"description"` instead of using the dynamically configured field name from `self._vector_search_fields["text_field"]`.

## Problem

When performing vector or hybrid searches without custom projection mapping, the code has two inconsistencies:

1. **Line 900** (`_generate_projection_fields` method): The SQL projection hardcodes the alias as `description`:
   ```python
   projection = f"{table}.id, {table}[@textKey] as description, ..."
   ```

2. **Line 1007** (`_execute_query` method): The fallback when `projection_mapping` is not provided uses the hardcoded string `"description"` instead of the configured field name:
   ```python
   text_key = "description"
   ```

This causes a mismatch between:
- The configured text field name in `vector_search_fields["text_field"]`
- The actual alias used in SQL queries
- The field name used to extract data from query results

## Impact

- Potential `KeyError` at runtime when the configured text field name differs from `"description"`
- Inconsistent behavior between different search types (full-text search works correctly)

## Changes
1. Updated line 900 to use the dynamic field name as alias
2. Updated line 1007 to use `self._vector_search_fields["text_field"]` instead of hardcoded `"description"`

### DIFF
--- a/.github/workflows/_compile_integration_test.yml
+++ b/.github/workflows/_compile_integration_test.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - "3.9"
+          - "3.10"
           - "3.12"
     name: "poetry run pytest -m compile tests/integration_tests #${{ matrix.python-version }}"
     steps:
--- a/.github/workflows/_lint.yml
+++ b/.github/workflows/_lint.yml
@@ -29,7 +29,7 @@ jobs:
         # Starting new jobs is also relatively slow,
         # so linting on fewer versions makes CI faster.
         python-version:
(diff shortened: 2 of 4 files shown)


davide-evangelisti-prom:
Hi @nagkumar91 ,
Sorry if you’re not the right person to ask — if so, feel free to point me in the right direction.
Whenever possible, could someone take a look at this PR?
Thanks!

PR #206, opened 2025-11-13
Demo for AzureBlobStorageLoader with RAG agent

anjaliratnam-msft:
Added a demo for using the AzureBlobStorageLoader in a RAG application.

### DIFF
--- a/samples/rag-storage-document-loaders/embed.py
+++ b/samples/rag-storage-document-loaders/embed.py
@@ -44,7 +44,7 @@ def main() -> None:
         azure_search_key=None,
         azure_credential=_CREDENTIAL,
         additional_search_client_options=_COGNITIVE_CREDENTIAL_SCOPES,
-        index_name=os.environ.get("AZURE_SEARCH_INDEX_NAME", "demo-documents"),
+        index_name=os.environ.get("AZURE_AI_SEARCH_INDEX_NAME", "demo-documents"),
         embedding_function=embed_model,
     )
 
--- a/samples/rag-storage-document-loaders/query.py
+++ b/samples/rag-storage-document-loaders/query.py
@@ -43,7 +43,7 @@ def get_azure_search() -> AzureSearch:
         azure_search_endpoint=os.environ["AZURE_AI_SEARCH_ENDPOINT"],
         azure_search_key=None,
         azure_credential=_CREDENTIAL,
-        index_name=os.environ.get("AZURE_SEARCH_INDEX_NAME", "demo-documents"),
+        index_name=os.environ.get("AZURE_AI_SEARCH_INDEX_NAME", "demo-documents"),
         embedding_function=embed_model,
         additional_search_client_options=_COGNITIVE_CREDENTIAL_SCOPES,
     )


libs/azure-storage/demos/document_loaders/embed.py
```diff
@@ -0,0 +1,58 @@
...
+    loader = AzureBlobStorageLoader(
+        account_url=os.environ["AZURE_STORAGE_ACCOUNT_URL"],
+        container_name=os.environ["AZURE_STORAGE_CONTAINER_NAME"],
+        blob_names=[
+            "pdf_test.pdf",
+            "pdf_file2.pdf",
+            "pdf_file3.pdf",
+            "pdf_file4.pdf",
+            "pdf_file5.pdf",
+        ],
```
kyleknap: It would be interesting if we could update the loader to support listing from the container itself to show how it can automatically find blobs. I'm thinking we can just set the `prefix` and read from an `AZURE_STORAGE_BLOB_PREFIX` env var that is optional and set the value to `None` if it is not set.

libs/azure-storage/demos/document_loaders/requirements.txt
```diff
@@ -0,0 +1,5 @@
+-e ../..
```
kyleknap: Let's install directly from the PyPI package `langchain-azure-storage` instead so that this can be used standalone of the repository structure.

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,58 @@
...
+import os
+import warnings
+
+from azure.identity import DefaultAzureCredential
+from langchain_azure_ai.embeddings import AzureAIEmbeddingsModel
+from langchain_azure_ai.vectorstores import AzureSearch
+from langchain_community.document_loaders import PyPDFLoader
+
+from langchain_azure_storage.document_loaders import AzureBlobStorageLoader
+
```
kyleknap, line 14: In both of the scripts, I think we need to actually call `load_dotenv()` per the [documentation](https://github.com/theskumar/python-dotenv?tab=readme-ov-file#getting-started) to actually get these environment variables loaded.

libs/azure-storage/demos/document_loaders/demo.py
```diff
@@ -0,0 +1,101 @@
...
+    while True:
+        user_input = input("\n\nYou: ")
+        if user_input.lower() == "exit":
+            print("\nGoodbye!")
+            break
+        try:
+            response = get_response(user_input, retriever, llm)
+            print(f"\nAI: {response}")
+        except Exception as e:
+            print(f"Error processing your request: {e}")
```
kyleknap: Let's remove the try/except to avoid swallowing errors/tracebacks. It will also help simplify the code.

libs/azure-storage/demos/document_loaders/demo.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def chatbot() -> None:
+    """Main chatbot loop."""
+    retriever = create_retriever()
+    llm = get_chat_model()
+    print("Welcome! Type 'exit' to quit.")
+
+    while True:
+        user_input = input("\n\nYou: ")
```
kyleknap: It would be interesting to give a little more context on what the chatbot will do as part of the welcome. Maybe include details that it will query previously embedded content from a vector store and it will answer questions based on that content.

libs/azure-storage/demos/document_loaders/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+3. **Create vector store** (first time only):
+   This step will load documents from Azure Blob Storage and save it to the Azure AI Search vector store.
+   ```bash
+   python embed.py
+   ```
+
+4. **Run the agent:**
+    ```bash
+    python demo.py
```
kyleknap: Maybe let's rename this script to something more self-descriptive like `query.py` or `retrieve.py`? Mainly it is not too obvious what we are running based on the name

libs/azure-storage/demos/document_loaders/embed.py
```diff
@@ -0,0 +1,58 @@
...
+        loader_factory=PyPDFLoader,
```
kyleknap: It would be good to add a comment here to indicate what the purpose of setting this loader. Mainly it will help in walking through the code

libs/azure-storage/demos/document_loaders/embed.py
```diff
@@ -0,0 +1,58 @@
...
+        azure_credential=DefaultAzureCredential(),
+        additional_search_client_options={
+            "credential_scopes": ["https://cognitiveservices.azure.com/.default"]
+        },
+        index_name="demo-documents",
+        embedding_function=embed_model,
+    )
+
+    for doc in loader.lazy_load():
+        azure_search.add_documents([doc])
```
kyleknap: It would be interesting if we could update the `add_documents()` call to add multiple documents at a time. I suspect that will help make this run faster especially when we add more documents to the index.

libs/azure-storage/demos/document_loaders/demo.py
```diff
@@ -0,0 +1,101 @@
...
+import os
+import warnings
+
+from azure.identity import DefaultAzureCredential
+from langchain_azure_ai.chat_models import AzureAIChatCompletionsModel
+from langchain_azure_ai.embeddings import AzureAIEmbeddingsModel
+from langchain_azure_ai.vectorstores import AzureSearch
+from langchain_community.vectorstores.azuresearch import AzureSearchVectorStoreRetriever
+
+warnings.filterwarnings("ignore")
```
kyleknap: We should circle back to update these filters to be specific to the warnings we are trying to ignore. Mainly it is a best practice to not suppress all warnings, only the warnings that we care about supressing.

kyleknap (commented):
This is looking good! Just left some comments based on the first pass.

samples/rag-storage-document-loaders/README.md
```diff
@@ -0,0 +1,39 @@
...
+   ```
+
+2. **Configure environment variables:**
+   ```bash
+   dotenv set AZURE_STORAGE_ACCOUNT_URL "https://<your-account-name>.blob.core.windows.net"
+   dotenv set AZURE_STORAGE_CONTAINER_NAME "your-container-name"
+   dotenv set AZURE_STORAGE_BLOB_PREFIX "your-blob-prefix"  # Defaults to `None`
+   dotenv set AZURE_EMBEDDING_ENDPOINT "your-azure-embedding-endpoint"
+   dotenv set AZURE_CHAT_ENDPOINT "your-azure-chat-endpoint"
+   dotenv set AZURE_AI_SEARCH_ENDPOINT "your-azure-search-service-endpoint"
```
kyleknap: It would be nice if we can add the placeholders like we did for the account url to make it clear what a value looks like. Otherwise, I'd suspect some friction figuring out what the sort of value being requested here.

samples/rag-storage-document-loaders/README.md
```diff
@@ -0,0 +1,39 @@
...
+   ```
+
+3. **Create vector store** (first time only):
+   This step will load documents from Azure Blob Storage and save it to the Azure AI Search vector store.
```
kyleknap: It would be nice to explain that this will list blobs as documents from the container and you can control blobs enumerated over by setting the blob prefix env var.

samples/rag-storage-document-loaders/README.md
```diff
@@ -0,0 +1,39 @@
...
+   ```bash
+   python embed.py
+   ```
+
+4. **Run the agent:**
```
kyleknap, line 32: Let's expand this description to call out that it retrieves context from the Azure AI Search vector store when answering requests.

samples/rag-storage-document-loaders/query.py
```diff
@@ -0,0 +1,105 @@
...
+    return response.content
+
+
+def chatbot() -> None:
+    """Main chatbot loop."""
+    retriever = create_retriever()
+    llm = get_chat_model()
+    print(
+        "Welcome! This chatbot answers questions based on indexed documents "
+        "stored in a vector store. Type 'exit' to quit."
```
kyleknap: More of a usability thing here. Instead of needing to type `exit` can we just have it be hit `<ENTER>` without any text input? That is a more common pattern for terminal based apps and if you do hit enter as of right now, it causes the app to crash with not an unclear exception message related to serializing the API request.

samples/rag-storage-document-loaders/query.py
```diff
@@ -0,0 +1,105 @@
...
+load_dotenv()
+warnings.filterwarnings("ignore", message=".*is currently in preview.*")
+
+
+def get_chat_model() -> AzureAIChatCompletionsModel:
+    """Initialize and return the Azure AI Chat Completions Model."""
+    chat_model = AzureAIChatCompletionsModel(
+        endpoint=os.environ["AZURE_CHAT_ENDPOINT"],
+        credential=DefaultAzureCredential(),
+        model="gpt-4.1-mini",
```
kyleknap: For both the chat model and embeddings model, could we expose these as separate required env vars? Mainly it will allow a customer to use different models without changing the code.

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+    azure_search = AzureSearch(
+        azure_search_endpoint=os.environ["AZURE_AI_SEARCH_ENDPOINT"],
+        azure_search_key=None,
+        azure_credential=DefaultAzureCredential(),
+        additional_search_client_options={
+            "credential_scopes": ["https://cognitiveservices.azure.com/.default"]
+        },
+        index_name="demo-documents",
```
kyleknap: We should also expose this as an environment variable, but we can have it default to `demo-documents` to make it less steps to setup.

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,64 @@
...
+        credential=DefaultAzureCredential(),
+        model="text-embedding-3-large",
+        client_kwargs={
+            "credential_scopes": ["https://cognitiveservices.azure.com/.default"]
+        },
+    )
+
+    azure_search = AzureSearch(
+        azure_search_endpoint=os.environ["AZURE_AI_SEARCH_ENDPOINT"],
+        azure_search_key=None,
```
kyleknap, line 44: Can we remove the `azure_search_key` parameter?

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,64 @@
...
+from langchain_azure_ai.embeddings import AzureAIEmbeddingsModel
+from langchain_azure_ai.vectorstores import AzureSearch
+from langchain_community.document_loaders import PyPDFLoader
+
+from langchain_azure_storage.document_loaders import AzureBlobStorageLoader
+
+
+load_dotenv()
+warnings.filterwarnings("ignore", message=".*is currently in preview.*")
+warnings.filterwarnings("ignore", message=".*`AzureBlobStorageLoader` is in public preview.*")
```
kyleknap: For both the `embed.py` and `query.py` lets maybe just use a message of:
```
".* preview.*
```
to just simplify this a bit but still only target preview related warnings.

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,64 @@
...
+        loader_factory=PyPDFLoader,  # Parses blobs as PDFs into LangChain Documents
+    )
+
+    embed_model = AzureAIEmbeddingsModel(
+        endpoint=os.environ["AZURE_EMBEDDING_ENDPOINT"],
+        credential=DefaultAzureCredential(),
+        model="text-embedding-3-large",
+        client_kwargs={
+            "credential_scopes": ["https://cognitiveservices.azure.com/.default"]
+        },
```
kyleknap: It might be worth declaring these kwargs as a module-level constant in both of these scripts to help cut down in the verbosity:
```python
_COGNITIVE_CREDENTIAL_SCOPES = {
      "credential_scopes": ["https://cognitiveservices.azure.com/.default"]
}
```

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,64 @@
...
+    loader = AzureBlobStorageLoader(
+        account_url=os.environ["AZURE_STORAGE_ACCOUNT_URL"],
+        container_name=os.environ["AZURE_STORAGE_CONTAINER_NAME"],
+        prefix=os.environ.get("AZURE_STORAGE_BLOB_PREFIX", None),
+        loader_factory=PyPDFLoader,  # Parses blobs as PDFs into LangChain Documents
+    )
+
+    embed_model = AzureAIEmbeddingsModel(
+        endpoint=os.environ["AZURE_EMBEDDING_ENDPOINT"],
+        credential=DefaultAzureCredential(),
```
kyleknap: A little unfortunate that it won't allow us to show off the default azure credential but from a best practices perspective, it probably makes sense to create a single instance of the default credentials and provide it to each client instead of instantiate new ones. Mainly in reusing the credential object there is caching that we can leverage for overall perf improvements.

samples/rag-storage-document-loaders/README.md
```diff
@@ -0,0 +1,39 @@
...
+    ```bash
+    python query.py
+    ```
+
+   **Sample interaction:**
+    ```text
+    You: What is Azure Blob Storage?
+
+    AI: Azure Blob Storage is a service for storing large amounts of unstructured data...
+    Source:  https://<your-account-name>.blob.core.windows.net/documents/pdf_file.pdf
```
kyleknap: Let's also swap `documents` to include the placeholder for container name to indicate it matches the value set in the env var section

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,64 @@
...
+        additional_search_client_options={
+            "credential_scopes": ["https://cognitiveservices.azure.com/.default"]
+        },
+        index_name="demo-documents",
+        embedding_function=embed_model,
+    )
+
+    batch = []
+    batch_size = 50
+    for doc in loader.lazy_load():
```
kyleknap: I wonder if we can use [`itertools.batched()`](https://docs.python.org/3/library/itertools.html#itertools.batched) here instead? It's only available in python 3.12 but that should be fine given this is just a sample demo. It would help clean this logic up a bit.

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,64 @@
...
+        azure_credential=DefaultAzureCredential(),
+        additional_search_client_options={
+            "credential_scopes": ["https://cognitiveservices.azure.com/.default"]
+        },
+        index_name="demo-documents",
+        embedding_function=embed_model,
+    )
+
+    batch = []
+    batch_size = 50
```
kyleknap: We should declare this as a module-level constant (e.g., `_EMBED_BATCH_SIZE`) to help clear up the main body of the code and make it clear this is more of a configuration

samples/rag-storage-document-loaders/embed.py
```diff
@@ -0,0 +1,58 @@
...
+        model=os.environ["AZURE_EMBEDDING_MODEL"],
+        client_kwargs=_COGNITIVE_CREDENTIAL_SCOPES.copy(),
+    )
+
+    azure_search = AzureSearch(
+        azure_search_endpoint=os.environ["AZURE_AI_SEARCH_ENDPOINT"],
+        azure_search_key=None,
+        azure_credential=_CREDENTIAL,
+        additional_search_client_options=_COGNITIVE_CREDENTIAL_SCOPES,
+        index_name=os.environ.get("AZURE_SEARCH_INDEX_NAME", "demo-documents"),
```
kyleknap: I think the environment variable name here should be `AZURE_AI_SEARCH_INDEX_NAME` here per the readme.

kyleknap (commented):
Looks good! Just left one comment that I ran into when working through it end to end. Otherwise, we should be set to merge it.

PR #211, opened 2025-11-15
AzureBlobStorageLoader demo updates

anjaliratnam-msft:
Updated the AzureBlobStorageLoader demo.

### DIFF
--- a/samples/rag-storage-document-loaders/embed.py
+++ b/samples/rag-storage-document-loaders/embed.py
@@ -35,6 +35,11 @@ def main() -> None:
         loader_factory=PyPDFLoader,  # Parses blobs as PDFs into LangChain Documents
     )
 
+    text_splitter = RecursiveCharacterTextSplitter(
+        chunk_size=12000,
+        chunk_overlap=500,
+    )
+
     embed_model = AzureAIEmbeddingsModel(
         endpoint=os.environ["AZURE_EMBEDDING_ENDPOINT"],
         credential=_CREDENTIAL,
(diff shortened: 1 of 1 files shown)


samples/rag-storage-document-loaders/embed.py
```diff
@@ -48,8 +51,22 @@ def main() -> None:
         embedding_function=embed_model,
     )
 
-    for batch in batched(loader.lazy_load(), _EMBED_BATCH_SIZE):
-        azure_search.add_documents(list(batch))
+    text_splitter = RecursiveCharacterTextSplitter(
+        chunk_size=12000,
+        chunk_overlap=500,
+        length_function=len,
```
kyleknap: Let's remove the `length_function` kwarg. It looks like `len` is the default value.

samples/rag-storage-document-loaders/embed.py
```diff
@@ -48,8 +51,22 @@ def main() -> None:
         embedding_function=embed_model,
     )
 
-    for batch in batched(loader.lazy_load(), _EMBED_BATCH_SIZE):
-        azure_search.add_documents(list(batch))
+    text_splitter = RecursiveCharacterTextSplitter(
```
kyleknap: Let's move the recursive text splitter declaration to in between the loader and the embeddings model to better represent where it is used in this pipeline.

samples/rag-storage-document-loaders/embed.py
```diff
@@ -48,8 +51,22 @@ def main() -> None:
...
+        splits = text_splitter.split_documents([doc])
+        docs.extend(splits)
+        if len(docs) >= _EMBED_BATCH_SIZE:
+            print(
+                f"Embedding and adding batch of {len(docs)} documents to Azure Search..."
+            )
+            azure_search.add_documents(list(docs))
+            docs = []
 
     print("Documents embedded and added to Azure Search index.")
```
kyleknap: For the progress, it would be nice to say how many docs we have added and how many blobs we have worked on. Otherwise, it is not easy to track. Something like this was working well for me:
```python
    docs = []
    total_processed = 0
    blobs_seen = set()
    for doc in loader.lazy_load():
        blobs_seen.add(doc.metadata.get("source"))
        splits = text_splitter.split_documents([doc])
        docs.extend(splits)
        if len(docs) >= _EMBED_BATCH_SIZE:
            azure_search.add_documents(docs)
            total_processed += len(docs)
            print(f"\rAdded {total_processed} total documents across {len(blobs_seen)} total blobs", end="\r", flush=True)
            docs = []

    print(f"Complete: {total_processed} documents across {len(blobs_seen)} blobs embedded and added to Azure Search index.")
```

kyleknap (approved):
Looks good! :ship:

PR #214, opened 2025-11-19
AzureBlobStorageLoader sample updates

anjaliratnam-msft:
Updated the AzureBlobStorageLoader sample.

### DIFF
--- a/samples/rag-storage-document-loaders/embed.py
+++ b/samples/rag-storage-document-loaders/embed.py
@@ -61,14 +61,9 @@ def main() -> None:
     docs = []
     total_processed = 0
     blobs_seen = set()
-    total_blobs = get_num_blobs()
-    blob_progress = tqdm(total=total_blobs, desc="Processing blobs", unit=" blobs")
+    blob_progress = get_progress_bar()
     for doc in loader.lazy_load():
-        blob = doc.metadata.get("source")
-        if blob not in blobs_seen:
-            blob_progress.update(1)
-
-        blobs_seen.add(blob)
+        update_progress_bar(doc, blobs_seen, blob_progress)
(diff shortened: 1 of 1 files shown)


samples/rag-storage-document-loaders/embed.py
```diff
@@ -59,24 +61,43 @@ def main() -> None:
...
+        azure_search.add_documents(docs)
+        total_processed += len(docs)
+
+    blob_progress.close()
     print(
         f"Complete: {total_processed} documents across {len(blobs_seen)} blobs embedded and added to Azure Search index."
     )
 
 
+def get_num_blobs() -> int:
```
kyleknap: Let's just repurpose this helper function to retrieve a progress bar e.g., `get_progress_bar()` to help declutter some of the main logic and make the embedding wiring easier to follow.

samples/rag-storage-document-loaders/embed.py
```diff
@@ -59,24 +61,43 @@ def main() -> None:
...
     blobs_seen = set()
+    total_blobs = get_num_blobs()
+    blob_progress = tqdm(total=total_blobs, desc="Processing blobs", unit=" blobs")
     for doc in loader.lazy_load():
-        blobs_seen.add(doc.metadata.get("source"))
+        blob = doc.metadata.get("source")
+        if blob not in blobs_seen:
+            blob_progress.update(1)
+
+        blobs_seen.add(blob)
```
kyleknap: For this logic, it may make sense to turn this into an `update_progress_bar()` function that accepts the current document, blobs we've seen, and the progress bar. This should help reduce some of the verbosity here in reading the main logic of the script

PR #231, opened 2025-12-11
Add a parameter to control whether remove image base64 data

ShichaoQiu:
fix issue #113 

To keep the base64 data of image, here is some example commands
```python
tool.run({"python_code": CODE, "remove_image_base64": False})
tool.invoke({"python_code": CODE, "remove_image_base64": False})
tool.ainvoke({"python_code": CODE, "remove_image_base64": False})
```

### DIFF
--- a/libs/azure-dynamic-sessions/pyproject.toml
+++ b/libs/azure-dynamic-sessions/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "langchain-azure-dynamic-sessions"
-version = "0.4.0"
+version = "0.3.1"
 description = "An integration package connecting Azure Container Apps dynamic sessions and LangChain"
 license = { file = "LICENSE" }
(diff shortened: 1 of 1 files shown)


libs/azure-dynamic-sessions/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "langchain-azure-dynamic-sessions"
-version = "0.3.0"
+version = "0.4.0"
```
santiagxf: Since this is a bug fix, can we use a minor version? `0.3.1`?

santiagxf: Since this is a bug fix, let's use a minor version.

```suggestion
version = "0.3.1"
```

santiagxf (approved):

PR #314, opened 2026-03-03
fix: prevent duplicate span export in AzureAIOpenTelemetryTracer

### LINKED ISSUE #310: Duplicate Spans in AzureAIOpenTelemetryTracer
## Summary
Every span created by `AzureAIOpenTelemetryTracer` appears **twice** in Azure Application Insights with identical span IDs, timestamps, and durations.

## Environment
- **Library**: `langchain-azure-ai>=0.1.0`
- **Azure Monitor**: `azure-monitor-opentelemetry>=1.0.0`
- **OpenTelemetry SDK**: `opentelemetry-sdk>=1.20.0`
- **Python**: 3.11+
- **LangChain**: 0.3.0+
- **LangGraph**: 1.0.0+

### Setup
```python
from langchain_azure_ai.callbacks.tracers import AzureAIOpenTelemetryTracer

# Initialize tracer
tracer = AzureAIOpenTelemetryTracer(
    connection_string="InstrumentationKey=...",
    enable_content_recording=True,
    provider_name="aws.bedrock",  # or "anthropic"
)
```

### Code
```python
from langchain_aws import ChatBedrock
from langchain_core.messages import HumanMessage

# Create LLM
llm = ChatBedrock(
    model_id="anthropic.claude-3-haiku-20240307-v1:0",
    temperature=0.2,
)

# Invoke with tracer callback
config = {
    "callbacks": [tracer],
    "metadata": {
        "agent_name": "Test Agent",
        "gen_ai.operation.name": "chat",
    }
}

response = llm.invoke([HumanMessage(content="Hello")], config=config)
```

### Query Azure Application Insights
```kusto
dependencies
| where timestamp > ago(5m)
| where name contains "chat"
| project timestamp, name, operation_Id, id, duration
| order by timestamp desc
```

### Expected Result
One span per LLM call.

### Actual Result
**Two identical spans** with:
- Same `operation_Id`
- Same span `id`
- Same `timestamp`
- Same `duration`

Example:
```
timestamp                   | name | operation_Id               | id               | duration
2026-03-02T06:50:03.798Z   | chat | abc-123                    | span-xyz         | 3236ms
2026-03-02T06:50:03.798Z   | chat | abc-123                    | span-xyz         | 3236ms
```

### Root Cause Theory
Looking at the source code, `AzureAIOpenTelemetryTracer._configure_azure_monitor()`:

```python
@classmethod
def _configure_azure_monitor(cls, connection_string: str) -> None:
    with cls._configure_lock:
        if cls._azure_monitor_configured:
            return
        configure_azure_monitor(connection_string=connection_string)
        cls._azure_monitor_configured = True
```

**Hypothesis**: The `configure_azure_monitor()` call creates a global TracerProvider with Azure Monitor exporters, but the tracer callback ALSO creates spans that get exported. This results in spans being exported twice:
1. Once by the global TracerProvider set up by `configure_azure_monitor()`
2. Once by the tracer callback mechanism

### Additional Evidence
- Issue occurs with both `ChatBedrock` (AWS Bedrock) and `ChatAnthropic` (Anthropic)
- Issue occurs in both AWS Lambda and local development
- `invoke_agent` spans also duplicated
- All span types affected: `chat`, `POST`, `invoke_agent`

## Impact
- **Data volume**: 2x telemetry data → higher costs
- **Query accuracy**: Counts/aggregations are 2x the actual value
- **Trace analysis**: Confusing to see duplicate spans in traces
- **Dashboards**: Metrics show double the actual numbers

### 1. Type = "N/A" for LLM spans
`chat` and `POST` spans show type="N/A" instead of expected "GenAI | <provider>"

Expected: `type = "GenAI | aws.bedrock"`
Actual: `type = "N/A"`

Note: `invoke_agent` spans correctly show `type = "InProc"`

### 2. Gen AI Attributes Not Queryable
Cannot query Gen AI semantic convention attributes from App Insights API:

```kusto
dependencies
| extend agent_name = tostring(customDimensions.['gen_ai.agent.name'])
| where isnotempty(agent_name)
```

Returns: `No rows returned`

Possible causes:
- Attributes not being set
- Attributes filtered by exporter
- Stored in different location

## Workarounds Attempted
None successful. The duplicate issue persists regardless of:
- Removing custom OTel TracerProvider setup
- Using ONLY `AzureAIOpenTelemetryTracer` callbacks
- Different LLM providers (Bedrock vs Anthropic)

## Questions
1. Is `configure_azure_monitor()` supposed to be called by the tracer, or should users call it separately?
2. Should we create a TracerProvider manually or rely on the tracer to do it?
3. Is there a recommended way to use `AzureAIOpenTelemetryTracer` to avoid duplicates?

## Related
- OpenTelemetry Gen AI Semantic Conventions: https://opentelemetry.io/docs/specs/semconv/gen-ai/
- Azure Monitor OpenTelemetry: https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-enable

nagkumar91:
## Summary
Fixes #310 — duplicate spans in Azure Application Insights from `AzureAIOpenTelemetryTracer`.

## Problem
Every span created by the tracer appears **twice** in Application Insights with identical span IDs, timestamps, and durations. This doubles telemetry costs and skews metrics.

## Root Cause
Three issues in `_configure_azure_monitor()`:

1. **No detection of existing TracerProvider** — if the user (or an auto-instrumentation agent on App Service/AKS/Lambda) already configured Azure Monitor, the tracer's call to `configure_azure_monitor()` creates a second configuration. While `set_tracer_provider()` only succeeds once, the second `configure_azure_monitor()` call still runs `_setup_instrumentations()` which re-instruments HTTP libraries, and creates an orphaned `TracerProvider` with its own `BatchSpanProcessor`.

2. **HTTP auto-instrumentation creates extra spans** — `configure_azure_monitor()` by default instruments `requests`, `urllib`, `urllib3`. These create HTTP `POST` spans alongside the tracer's own `chat` spans — perceived as duplicates.

3. **No opt-out mechanism** — users who manage their own `TracerProvider` had no way to prevent the tracer from calling `configure_azure_monitor()`.

## Changes
Three minimal, backward-compatible changes to `inference_tracing.py`:

1. **Detect existing TracerProvider**: Before calling `configure_azure_monitor()`, check if a real (non-`ProxyTracerProvider`) TracerProvider already exists. If so, skip the call and log an info message.

2. **Disable HTTP auto-instrumentors**: Pass `instrumentation_options` to `configure_azure_monitor()` disabling `requests`/`urllib`/`urllib3` — the tracer already creates its own GenAI spans for LLM calls.

3. **Add `auto_configure_azure_monitor` parameter**: New `__init__` kwarg (default `True` for backward compat) that lets users explicitly skip auto-configuration:
   ```python
   tracer = AzureAIOpenTelemetryTracer(
       connection_string="...",
       auto_configure_azure_monitor=False,  # I manage my own TracerProvider
   )
   ```

Also reorders `__init__` to call `_configure_azure_monitor()` **before** `otel_trace.get_tracer()` so the tracer is obtained from the real provider rather than through a ProxyTracer.

## Answers to Issue Questions

**Q1: Is `configure_azure_monitor()` supposed to be called by the tracer, or should users call it separately?**
The tracer calls it as a convenience when a `connection_string` is provided. Users should do **one or the other**, not both. After this fix, the tracer detects if a provider is already configured and skips. Users who configure it themselves can also pass `auto_configure_azure_monitor=False`.

**Q2: Should we create a TracerProvider manually or rely on the tracer to do it?**
For simple setups, let the tracer handle it (the default). For advanced setups (custom sampling, multiple exporters, OTLP collectors), configure the TracerProvider yourself *before* creating the tracer, and pass `auto_configure_azure_monitor=False`.

**Q3: Is there a recommended way to use `AzureAIOpenTelemetryTracer` to avoid duplicates?**
After this fix, the tracer auto-detects and avoids duplicates. Until then: ensure `configure_azure_monitor()` is called **exactly once** — either by the tracer OR by your own code. If you have App Service/AKS auto-instrumentation enabled, disable it or use `auto_configure_azure_monitor=False`.

## Reproduction Status
⚠️ **Fix was developed through code analysis** — I did not reproduce the actual duplicate spans in Application Insights. The PR includes a [reproduction plan](https://github.com/langchain-ai/langchain-azure/issues/310) covering both local development and AWS Lambda scenarios that should be validated.

## Tests
- 4 new unit tests added covering: singleton guard, existing-provider detection, opt-out parameter, HTTP instrumentor disabling
- All 97 unit tests pass


libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
```diff
@@ -1111,11 +1111,17 @@ def __init__(
         trace_all_langgraph_nodes: bool = False,
         ignore_start_node: bool = True,
         compat_create_agent_filtering: bool = True,
+        auto_configure_azure_monitor: bool = True,
         _prepare_messages_fn: Optional[
             Callable[..., tuple[Optional[str], Optional[str]]]
```
santiagxf, line 1115: @nagkumar91 can we document parameters in docstring?

nagkumar91: I will. This PR is not needed for our release yet. It happens only when `configure_azure_monitor` is invoked before the langchain is invoked. I think we can wait until another customer runs into this issue.

PR #382, opened 2026-03-14
feat(azure-ai): emit OpenTelemetry GenAI client metrics from the Azure AI tracer

nagkumar91:
## Summary

The `langchain-azure-ai` tracer emitted GenAI spans but not the corresponding OpenTelemetry GenAI client metrics. This updates the tracer to publish spec-aligned metrics for token usage, request duration, and streaming chunk timing from the same model invocation lifecycle it already traces.

## Metric coverage

| Metric | Description |
|--------|-------------|
| `gen_ai.client.token.usage` | Number of input and output tokens used |
| `gen_ai.client.operation.duration` | GenAI operation duration |
| `gen_ai.client.operation.time_to_first_chunk` | Time to receive first chunk from streaming response |
| `gen_ai.client.operation.time_per_output_chunk` | Time between streaming output chunks after first chunk |

## Tracer integration

- Initializes an OpenTelemetry meter alongside the existing tracer
- Reuses existing span attributes to populate metric dimensions:
  - `gen_ai.provider.name`, `gen_ai.operation.name`
  - `gen_ai.request.model`, `gen_ai.response.model`
  - `server.address`, `server.port`
  - `error.type`, `gen_ai.token.type`

## Model lifecycle changes

- Records token usage metrics from normalized usage data in `on_llm_end`
- Records operation duration when model spans close
- Ensures failed model calls include `error.type` on emitted duration metrics

## Streaming behavior

- Tracks first streamed chunk latency per run via `on_llm_new_token`
- Tracks inter-chunk latency for subsequent output chunks
- Preserves current callback behavior (no span event mutation)

## Review-driven refinements

- `_MODEL_OPERATIONS` extracted as module-level `frozenset` to avoid per-call set allocation
- Base metric attributes cached per span for streaming hot paths
- Cache is invalidated when `gen_ai.response.model` becomes available so end-of-run metrics include refreshed dimensions
- `started_at` is now recorded only for model operations to avoid `perf_counter()` overhead on non-model spans
- Added regression coverage for streaming cache refresh behavior

## Test coverage

Extends tracer unit tests with coverage for:
- Token usage metrics
- Operation duration metrics
- First-chunk timing
- Per-chunk timing
- Error-path metric attributes
- Streaming cache invalidation when response model is learned at end-of-run

All targeted tests pass. Ruff + mypy clean.

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
+++ b/libs/azure-ai/langchain_azure_ai/callbacks/tracers/inference_tracing.py
@@ -1980,6 +1980,7 @@ class AzureAIOpenTelemetryTracer(BaseCallbackHandler):
         if "model_name" in llm_output:
             record.span.set_attribute(Attrs.RESPONSE_MODEL, llm_output["model_name"])
             record.attributes[Attrs.RESPONSE_MODEL] = llm_output["model_name"]
+            record.stash.pop("_metric_attrs_cache", None)
         if llm_output.get("system_fingerprint"):
             record.span.set_attribute(
                 Attrs.OPENAI_RESPONSE_SYSTEM_FINGERPRINT,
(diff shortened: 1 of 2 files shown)


nagkumar91:
Pushed ruff formatting fix in b8c1846. All 3 original review comments were already addressed in the previous commit (872d246). All review threads should now be outdated.

PR #457, opened 2026-04-07
fix(langchain-sqlserver): remove redundant session.close() inside con…

Serjbory:
### Summary

Removes an unnecessary `session.close()` call inside a `with Session(...)` block in `SQLServer_VectorStore._prepare_json_data_type()`.

### Problem

In `vectorstores.py`, `session.close()` is called explicitly inside a `with Session(self._bind) as session:` block:

```python
with Session(self._bind) as session:
    result = session.scalar(text(SERVER_JSON_CHECK_QUERY))
    session.close()  # <-- redundant

    if result is not None:
        ...
```

SQLAlchemy's `Session` context manager already calls `close()` in `__exit__`:

```python
# SQLAlchemy 2.0 source
def __exit__(self, type_, value, traceback):
    self.close()
```

This results in `close()` being called twice — once explicitly, then again when the `with` block exits. While `Session.close()` is idempotent (no crash), the explicit call is redundant and misleading.

### Fix

Remove the `session.close()` line, relying on the context manager to handle cleanup as intended.

### Verification

- Confirmed against SQLAlchemy 2.0.42 (the version resolved by `poetry.lock`)
- `make lint_package` passes (ruff + mypy)

### Package

`langchain-sqlserver`


Serjbory:
hello @beccadaniel ! Can you review this one, please or run copilot review (I will add changes if needed)?

libs/sqlserver/poetry.lock
beccadaniel, line 1: @Serjbory , this lock file appears to have been regenerated with a lower version. Can we change that?

Serjbory: @beccadaniel sure. I'll do it rn

Serjbory: ah, it's already was bumped when you merged. No actions needed from my side. Thanks !

beccadaniel:
@copilot resolve the merge conflicts in this pull request

Serjbory:
@beccadaniel it's done. Rebased on main and all conflicts are resolved

beccadaniel (changes requested):
@Serjbory, as mentioned earlier, the `poetry.lock` change isn't needed in the PR.

This change takes the poetry version to an earlier version and now fails CI. Let's remove the poetry file change so we can close on this PR.

Serjbory:
@beccadaniel, sure, it's done. Poetry file was removed from the PR. Only necessary changes are left

beccadaniel (approved):

PR #472, opened 2026-04-09
fix(azure-dynamic-sessions): move BaseSandbox import out of TYPE_CHEC…

### LINKED ISSUE #451: [langchain-azure-dynamic-sessions] SessionsBashBackend TYPE_CHECKING import error
Hi, I wanted to use `SessionsBashBackend` but it fails on import with `NameError: name 'BaseSandbox' is not defined`.

The import in `backends/sessions.py` is guarded by `TYPE_CHECKING`:

```python

from typing import TYPE_CHECKING, Callable, Optional

 

if TYPE_CHECKING:

    from deepagents.backends.protocol import (

        EditResult,

        ExecuteResponse,

        FileDownloadResponse,

        FileInfo,

        FileUploadResponse,

        WriteResult,

    )

    from deepagents.backends.sandbox import BaseSandbox

...

 

@experimental()

class SessionsBashBackend(BaseSandbox):

```

Is this intentional? If I delete this if statement it runs okay but with it it fails in runtime :<

Serjbory:
## fix(azure-dynamic-sessions): move `BaseSandbox` import out of `TYPE_CHECKING` guard

Fixes https://github.com/langchain-ai/langchain-azure/issues/451

### Problem

`SessionsBashBackend` in `langchain_azure_dynamic_sessions.backends.sessions` inherits from `BaseSandbox` (from `deepagents`), but the import is guarded by `if TYPE_CHECKING:`:

```python
if TYPE_CHECKING:
    from deepagents.backends.sandbox import BaseSandbox

class SessionsBashBackend(BaseSandbox):  # NameError at runtime
```

`TYPE_CHECKING` is always `False` at runtime, so `BaseSandbox` is never defined — making the class **impossible to import** regardless of whether `deepagents` is installed. While `from __future__ import annotations` makes type annotations lazy strings, base classes are evaluated eagerly at class-definition time.

### Fix

Move the `BaseSandbox` import out of the `TYPE_CHECKING` block into a `try/except ImportError` with an `object` fallback:

```python
try:
    from deepagents.backends.sandbox import BaseSandbox
except ImportError:
    BaseSandbox = object  # type: ignore[misc,assignment]
```

- **With `deepagents` installed** → imports the real `BaseSandbox`, class works as designed.
- **Without `deepagents`** → falls back to `object`, so the module is importable and doesn't crash the `backends/__init__.py` re-export.

The protocol types (`ExecuteResponse`, `EditResult`, etc.) remain under `TYPE_CHECKING` since they are only used in return-type annotations (which are strings at runtime thanks to `from __future__ import annotations`).

### Changes

| File | Description |
|------|-------------|
| `libs/azure-dynamic-sessions/langchain_azure_dynamic_sessions/backends/sessions.py` | Move `BaseSandbox` import to `try/except`, keep protocol types under `TYPE_CHECKING` |
| `libs/azure-dynamic-sessions/tests/unit_tests/test_imports.py` | Add `test_sessions_bash_backend_importable` |

### Verification

- All 21 unit tests pass (`make test`)
- Formatting clean (`make format`)
- Lint + mypy clean on both package and tests (`make lint_package`, `make lint_tests`)


santiagxf:
@Serjbory we fixed this on #473

PR #443, opened 2026-03-31
Adding langchain-azure-cosmosdb package

aayush3011:
# Add `langchain-azure-cosmosdb` — Unified Azure CosmosDB Package

## Description

This PR introduces a new standalone package **`langchain-azure-cosmosdb`** that consolidates all Azure CosmosDB NoSQL integrations for both LangChain and LangGraph into a single PyPI package. This enables customers to use one package (`pip install langchain-azure-cosmosdb`) to get all CosmosDB capabilities with LangChain and LangGraph.

The package will be maintained by the Microsoft Azure CosmosDB team.

## Acknowledgments

The CosmosDB checkpointer was originally developed by **[@skamalj (Kamal)](https://github.com/skamalj)** as a community project at [skamalj/langgraph_checkpoint_cosmosdb](https://github.com/skamalj/langgraph_checkpoint_cosmosdb). Per the LangChain team's guidance, the checkpointer has been moved here under `langchain-azure` so it can be released and maintained alongside other Azure integrations by the Microsoft CosmosDB team. The `langgraph-checkpoint-cosmosdb` package will be deprecated in favor of `langchain-azure-cosmosdb`.

**Huge thanks to [@skamalj (Kamal)](https://github.com/skamalj)** for creating the original checkpointer implementation. His community contribution is the foundation of the LangGraph checkpoint support in this package.

### LangChain Integrations (sync + async)

| Class | Description |
|-------|-------------|
| `AzureCosmosDBNoSqlVectorSearch` / `AsyncAzureCosmosDBNoSqlVectorSearch` | Vector store with vector, full-text, hybrid, and weighted hybrid search |
| `AzureCosmosDBNoSqlSemanticCache` / `AsyncAzureCosmosDBNoSqlSemanticCache` | LLM response caching backed by CosmosDB NoSQL |
| `CosmosDBChatMessageHistory` / `AsyncCosmosDBChatMessageHistory` | Persistent chat message history |

### LangGraph Integrations (sync + async)

| Class | Description |
|-------|-------------|
| `CosmosDBSaver` / `CosmosDBSaverSync` | LangGraph checkpoint persistence (async-first / sync) |
| `CosmosDBCache` / `CosmosDBCacheSync` | LangGraph cache with TTL support (async-first / sync) |

### All 13 classes are accessible from a single import:

```python
from langchain_azure_cosmosdb import (
    AzureCosmosDBNoSqlVectorSearch,
    AsyncAzureCosmosDBNoSqlVectorSearch,
[...]

PR #498, opened 2026-04-16
Add CosmosDB LangGraph BaseStore implementation (sync + async)

TheovanKraay:


libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+    grouped_ops: dict[type, list[tuple[int, Op]]] = defaultdict(list)
+    tot = 0
+    for idx, op in enumerate(ops):
+        grouped_ops[type(op)].append((idx, op))
+        tot += 1
+    return grouped_ops, tot
+
+
+def _ensure_index_config(
+    index_config: CosmosDBIndexConfig,
```
aayush3011, line 915: ### 🔴 Blocking — Namespace dot-separator causes data corruption

Namespace tuples are joined with `"."` and split on `"."`. If any namespace component contains a dot (e.g., `("users", "john.doe")`), the round-trip is lossy:
- Stored as prefix `"users.john.doe"`
- `_decode_ns` splits to `("users", "john", "doe")` — **3 elements instead of 2**
- Document IDs will also collide across logically distinct namespaces

LangGraph namespaces are user-controlled strings, so dots are a realistic scenario.

**The existing cache in this same repo** uses `"|"` as `_NS_SEPARATOR` (see `_langgraph_cache.py`) specifically to avoid this problem.

**Suggestion:** Use `"|"` as the separator (matching the cache convention), or a character like `\x00`. At minimum, validate components don't contain the separator and raise a clear error.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+        embedding: list[float] | None = None,
+    ) -> tuple[str, list[dict[str, Any]]]:
+        """Build a Cosmos DB query for search operations."""
+        params: list[dict[str, Any]] = []
+        conditions = []
+
+        # Namespace prefix filter
+        if op.namespace_prefix:
+            prefix = _namespace_to_text(op.namespace_prefix)
+            conditions.append("STARTSWITH(c.prefix, @ns_prefix)")
```
aayush3011: ### 🔴 Blocking — False positive namespace prefix matching in search

No trailing separator is appended. A search with `namespace_prefix=("test",)` generates `STARTSWITH(c.prefix, "test")`, which also matches prefix `"testing"` — returning items from namespace `("testing",)` that don't belong.

**Suggestion:** Use a compound condition:
```python
(c.prefix = @ns_prefix OR STARTSWITH(c.prefix, @ns_prefix_dot))
```
where `@ns_prefix_dot` = `"test."` (prefix + separator). This ensures exact namespace match OR proper sub-namespace match.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+        conditions = []
+
+        if op.match_conditions:
+            for condition in op.match_conditions:
+                if condition.match_type == "prefix":
+                    path = _namespace_to_text(
+                        tuple("%" if v == "*" else v for v in condition.path)
+                    )
+                    conditions.append("STARTSWITH(c.prefix, @match_prefix)")
+                    params.append({"name": "@match_prefix", "value": path})
```
aayush3011: ### 🔴 Blocking — Parameter name collision in list_namespaces

If multiple match conditions of the same type are passed, they all use the same hardcoded parameter name (`@match_prefix` or `@match_suffix`). The second one silently overwrites the first in the params list, producing incorrect query results.

**Suggestion:** Suffix parameter names with an index, consistent with how `_build_search_query` handles filter parameters:
```python
param_name = f"@match_prefix_{len(params)}"
```

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+        op: PutOp,
+        existing_doc: dict[str, Any] | None = None,
+    ) -> dict[str, Any]:
+        """Prepare a document for upsert."""
+        namespace = op.namespace
+        prefix = _namespace_to_text(namespace)
+        doc_id = self._make_doc_id(namespace, op.key)
+        now = datetime.now(UTC).isoformat()
+
+        created_at = now
```
aayush3011: ### 🟡 Significant — `created_at` overwritten on every update

`_prepare_put_document` accepts `existing_doc` to preserve `created_at`, but **no caller ever passes it**. Every call is `self._prepare_put_document(op)` with no second argument. This means `created_at == updated_at` on every upsert — the original creation timestamp is permanently lost on updates.

**Suggestion:** Either do a point-read before upsert to get the existing `created_at`, or remove the unused `existing_doc` parameter and document that `created_at` reflects the last write time (if that's the intended behavior).

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+            conn_string: Azure Cosmos DB connection string.
+            database_name: Name of the database to use.
+            container_name: Name of the container to use.
+            index: Optional index/embedding configuration for vector search.
+            ttl: Optional TTL configuration.
+
+        Returns:
+            A new CosmosDBStore instance.
+        """
+        client = CosmosClient.from_connection_string(conn_string)
```
aayush3011: ### 🟡 Significant — Missing User Agent string

Every other component in this repo sets a `USER_AGENT` constant and passes it to the Cosmos client. In PR #443, reviewer @santiagxf explicitly requested this for usage tracking, and it was added to all components as `langchain-azure-cosmosdb-<component>`.

**Suggestion:** Add to both sync and async:
```python
USER_AGENT = "langchain-azure-cosmosdb-lgstore"
client = CosmosClient.from_connection_string(conn_string, user_agent=USER_AGENT)
```

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+        # Compute embeddings for queries that need them
+        embedding_requests: list[tuple[int, str]] = []
+        for idx, (_, op) in enumerate(search_ops):
+            if op.query and self.index_config:
+                embedding_requests.append((idx, op.query))
+
+        embeddings: dict[int, list[float]] = {}
+        if embedding_requests and self.embeddings:
+            texts = [q for _, q in embedding_requests]
+            vectors = self.embeddings.embed_documents(texts)
```
aayush3011: ### 🟡 Significant — Search queries use `embed_documents` instead of `embed_query`

Search *queries* are embedded using `embed_documents`. The LangChain `Embeddings` interface separates `embed_query` from `embed_documents` because asymmetric embedding models (e.g., models that prepend "query:" vs "passage:" prefixes) produce different vectors for each purpose. Using `embed_documents` for queries will degrade search relevance with such models.

Same issue in `aio/_langgraph_store.py` — `_abatch_search_ops()` uses `aembed_documents` for queries.

**Suggestion:** Use `embed_query` / `aembed_query` for search queries. Keep `embed_documents` / `aembed_documents` for document indexing in put operations.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+
+        return query, params
+
+    def _get_filter_condition(
+        self, key: str, op: str, value: Any, param_offset: int
+    ) -> tuple[str, list[dict[str, Any]]]:
+        """Build a filter condition for Cosmos DB queries."""
+        param_name = f"@filter_{param_offset}"
+        if op == "$eq":
+            return f'c["value"]["{key}"] = {param_name}', [
```
aayush3011, line 235: ### 🟡 Significant — Filter key injection risk

Filter *values* are properly parameterized, but filter *keys* are interpolated directly into the SQL string without validation. Keys come from user input (`op.filter.keys()`). A malicious key like `"] OR 1=1 OR c["x` could alter query semantics.

The existing vectorstore in this repo has `_validate_sql_identifier()` for exactly this purpose.

**Suggestion:** Validate filter keys against a safe pattern (alphanumeric + underscore), or reuse `_validate_sql_identifier` from the vectorstore module.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+    Note:
+        Semantic search is disabled by default. Provide an ``index``
+        configuration when creating the store to enable it.
+
+    Warning:
+        Make sure to call ``setup()`` before first use to create the
+        database and container.
+    """
+
+    __slots__ = (
```
aayush3011, line 392: ### 🟡 Moderate — `ttl_config` missing from `__slots__`

`self.ttl_config = ttl` is assigned in `__init__`, but `ttl_config` is not listed here. This works currently because a parent class likely provides `__dict__`, but it's inconsistent with the declared intent of using `__slots__` and will break if the class hierarchy changes.

Same issue in `aio/_langgraph_store.py` `AsyncCosmosDBStore.__slots__`.

**Suggestion:** Add `"ttl_config"` to `__slots__` in both classes.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+                        "dimensions": dims,
+                    }
+                ]
+            }
+
+            container_kwargs["indexing_policy"] = {
+                "includedPaths": [{"path": "/*"}],
+                "excludedPaths": [{"path": "/embedding/*"}],
+                "vectorIndexes": [
+                    {"path": "/embedding", "type": "quantizedFlat"}
```
aayush3011: ### 🟡 Moderate — Vector index type hardcoded to `quantizedFlat`

For larger datasets (100K+ vectors), `diskANN` is significantly more performant. The index type is hardcoded with no way to override it.

**Suggestion:** Add an optional `index_type` field to `CosmosDBIndexConfig` defaulting to `"quantizedFlat"`, and use it here.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+                        ttl_seconds = int(float(default_ttl) * 60)
+                        doc["ttl"] = ttl_seconds
+                        doc["ttl_minutes"] = float(default_ttl)
+
+                # Add embedding if available
+                embedding = embeddings_map.get((op.namespace, op.key))
+                if embedding is not None:
+                    doc["embedding"] = embedding
+
+                prefix = _namespace_to_text(op.namespace)
```
aayush3011: ### 🟢 Nit — Dead code: unused `prefix` variable

`prefix` is computed here but never used. Looks like leftover from a refactoring — a reader might wonder if it was supposed to be passed as `partition_key` to `upsert_item`.

**Suggestion:** Remove this line.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+            tot += len(toks)
+    index_config["__tokenized_fields"] = tokenized
+    index_config["__estimated_num_vectors"] = tot
+    embeddings = ensure_embeddings(
+        index_config.get("embed"),
+    )
+    return embeddings, index_config
+
+
+__all__ = [
```
aayush3011, line 941: ### 🟢 Nit — `__all__` exports private functions

By convention, `__all__` defines the public API. `_decode_ns`, `_ensure_index_config`, `_group_ops`, `_namespace_to_text` are underscore-prefixed "private" functions. They're imported by the async module, but direct imports work regardless of `__all__`.

**Suggestion:** Remove the underscore-prefixed names from `__all__`.

libs/azure-cosmosdb/tests/integration_tests/aio/test_langgraph_store.py
```diff
@@ -0,0 +1,638 @@
+# type: ignore
```
aayush3011, line 1: ### 🟢 Nit — Async test coverage gap

The async test file covers ~40% fewer scenarios than the sync file. Notable missing tests:
- Search with operator filters (`$gt`, `$gte`, `$lt`, `$lte`, `$ne`)
- Search with offset/pagination
- List namespaces with prefix/suffix/max_depth
- Batch put with `value=None` (delete)
- Vector update/pagination/`index=False`/edge cases
- TTL expiry with actual sleep
- TTL fields verification in raw document

Since the async code has independently implemented query execution and embedding, async-specific bugs could hide in untested paths.

**Suggestion:** Port the key test classes to the async suite, at minimum: operator filters, namespace listing variants, and batch delete.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+                    params.append({"name": "@match_prefix", "value": path})
+                elif condition.match_type == "suffix":
+                    path = _namespace_to_text(
+                        tuple("%" if v == "*" else v for v in condition.path)
+                    )
+                    conditions.append("ENDSWITH(c.prefix, @match_suffix)")
+                    params.append({"name": "@match_suffix", "value": path})
+
+        where_clause = " AND ".join(conditions) if conditions else "1=1"
+        query = f"SELECT DISTINCT c.prefix FROM c WHERE {where_clause}"
```
aayush3011, line 225: ### 🟢 Nit — `list_namespaces` fetches all results into memory

`SELECT DISTINCT c.prefix` has no server-side limit. All matching prefixes are fetched into memory, deduplicated with `max_depth` truncation, then sliced with `offset`/`limit`. For stores with millions of documents, this pulls every distinct prefix before discarding most.

**Suggestion:** Add a generous server-side `TOP` clause as a safety cap, or document this as a known scalability limitation.

aayush3011 (changes requested):
## PR Review: CosmosDB LangGraph BaseStore Implementation

Overall the design is clean — the shared `BaseCosmosDBStore[C]` generic base avoids duplicating query logic, TTL is correctly delegated to Cosmos native TTL, and parameterized queries are used throughout. However, there are several correctness issues in namespace encoding and query construction that need addressing before merge.

**5 blocking issues** (data corruption, wrong query results), **6 significant/moderate** improvements, and **4 nits**.

See inline comments for details.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
```diff
@@ -0,0 +1,869 @@
...
+                if ns not in seen:
+                    seen.add(ns)
+                    namespaces.append(ns)
+
+            # Apply pagination
+            start = op.offset
+            end = start + op.limit
+            results[original_idx] = namespaces[start:end]
+
+    async def abatch(self, ops: Iterable[Op]) -> list[Result]:
```
aayush3011, line 820: What does this method do?

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: This matches the pattern in all LangGraph BaseStore implementations (PostgreSQL, InMemory, etc.) on which this was based.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Fixed. Changed separator from "." to "|".

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Fixed, replaced with compound condition.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Fixed. Hardcoded @match_prefix / @match_suffix now indexed (i.e. @match_prefix_{len(params)}, @match_suffix_{len(params)}). Each match condition gets a unique parameter name.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Fixed. Added a point-read before upsert to preserve the original created_at timestamp (1 RU). Changed existing_doc param to existing_created_at: str | None for clarity. Applied to both sync and async stores.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Added USER_AGENT = "langchain-azure-cosmosdb-lgstore" constant. Both from_conn_string() and from_endpoint() pass user_agent=USER_AGENT to the CosmosClient constructor.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Added index_type: Literal["quantizedFlat", "flat", "diskANN"] to CosmosDBIndexConfig. Setup reads self.index_config.get("index_type", "quantizedFlat") instead of hardcoding. Users can now specify diskANN for large-scale workloads.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Good catch, fixed. Search now calls embed_query()/aembed_query() per search operation instead of batching through embed_documents(). LangChain separates these methods to support asymmetric embedding models where query and document embeddings differ. Updated the AsyncCharacterEmbeddings test helper to track embed_query/aembed_query calls as well.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Added validate_filter_key() that validates filter keys against ^[A-Za-z_][A-Za-z0-9_]*$ before they're interpolated into the SQL query string, which matches the pattern used in _vectorstore.py.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Added "ttl_config" to __slots__ in both CosmosDBStore and AsyncCosmosDBStore. Without it, the attribute was being stored in a fallback __dict__ (since the base class doesn't use __slots__), but cleaner to declare it explicitly.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: removed

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Cleaned up.

libs/azure-cosmosdb/tests/integration_tests/aio/test_langgraph_store.py
TheovanKraay: Ack - increased coverage.

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: Also, NoSQL API does not natively support DISTINCT on array sub-paths or GROUP BY on dynamic-depth namespace components in a way that would let us do the prefix/suffix/max_depth filtering server-side. The current approach queries with SELECT DISTINCT c.namespace (which at least Cosmos does optimize) and then applies the filtering in Python. For stores with millions of distinct namespaces, could be an issue, but its an unusual case for a LangGraph store. Better to address this with backend optimization rather than over engineer here?

libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_store.py
TheovanKraay: The abatch function? Provides async compatible interface by delegating to the sync batch() method. The sync CosmosDBStore is built on the sync azure.cosmos.CosmosClient, so it can't natively do async I/O. Its abatch wraps batch() via run_in_executor as a compatibility shim so it satisfies LangChain's BaseStore interface (which requires abatch). Users who want true async should use AsyncCosmosDBStore (in aio), which is built on async azure.cosmos.aio.CosmosClient and has a fully native abatch. Done this way as separate classes because SDK has distinct sync/async clients that can't be shared. Added a docstring to CosmosDBStore.abatch to make this clear.

aayush3011:
@Copilot Fix the linting issues.

aayush3011 (approved):

PR #628, opened 2026-05-26
feat(sqlserver): add upsert option to add_texts and from_documents

### LINKED ISSUE #13: [langchain-sqlserver] Add support for upserting documents
When using `add_texts` or `add_documents` or `from_documents`, it should be possible to have an option named `upsert` that, if set to `true` will 
- insert a new row in the underlying table if a row with the same id doesn't exists or,
- update an existing row with the new data, if a row with the same id is existing already

The id used to check if a row is existing or not is the id provided by the developer, currently saved in the `custom_id` column, By default `upsert` will be set to `false` to keep consistency with other vector stores.

0xDevNinja:
## Summary

Adds an `upsert` keyword to `SQLServer_VectorStore.add_texts`, `from_texts` and `from_documents`. When `upsert=True`, rows whose `custom_id` already exists in the table are deleted before the new rows are inserted in the same transaction, so the call effectively updates them instead of raising on the unique-index conflict.

Default remains `upsert=False` to preserve the existing insert-only behavior consistent with other vector stores.

The delete + insert run in the same SQLAlchemy session/transaction, so an insert failure rolls the delete back — there is no window where the existing row is gone but the replacement has not landed.

Closes #13

## Changes

- `add_texts(..., upsert: bool = False)` — new keyword-only argument, forwarded to `_insert_embeddings`.
- `from_texts(..., upsert: bool = False)` and `from_documents(..., upsert: bool = False)` — same flag, forwarded to `add_texts`.
- `_insert_embeddings` — when `upsert=True`, deletes any rows whose `custom_id` matches an incoming row before issuing the bulk `INSERT`, inside the same transaction.

## Test plan

- [x] `make lint` clean on the changed files (`ruff check`, `mypy`)
- [x] Existing unit tests pass (`pytest tests/unit_tests/`)
- [ ] New integration tests run against a live SQL Server (require the `TEST_AZURESQLSERVER_TRUSTED_CONNECTION` env var, same setup as the existing integration suite):
  - `test_that_upsert_updates_existing_rows_by_custom_id` — overlapping ids are updated, untouched ids are preserved, no duplicates remain.
  - `test_that_upsert_default_false_still_raises_on_duplicate_custom_id` — default behavior is preserved.
  - `test_that_from_documents_supports_upsert` — flag is forwarded through `from_documents`.


santiagxf:
@0xDevNinja please let us know if you can address the issues with this PR.

0xDevNinja:
@santiagxf fixed this up — corrected the indentation in the upsert path, deduped the custom_id batch before the delete, and switched the test to a context-managed connection per the review notes. CI is green now.

PR #630, opened 2026-05-27
feat(sqlserver): add SQLServerSaver LangGraph checkpoint saver

### LINKED ISSUE #15: [langchain-sqlserver] Add support for storing chat history using LangGraph
The latest version (0.3) of LangChain mentions that

> As of the v0.3 release of LangChain, we recommend that LangChain users take advantage of [LangGraph persistence](https://langchain-ai.github.io/langgraph/concepts/persistence/) to incorporate memory into new LangChain applications.

So we should add support to [LangGraph persistence](https://langchain-ai.github.io/langgraph/concepts/persistence/) API to make the SQL Server LangChain library compatible also with the more powerful LangGraph framework

0xDevNinja:
## Summary

Adds `SQLServerSaver`, an implementation of `langgraph.checkpoint.base.BaseCheckpointSaver` backed by a SQL Server / Azure SQL database. This is the LangGraph-persistence integration tracked in #15.

Modeled after `langgraph.checkpoint.sqlite.SqliteSaver` so behavior is consistent with the other first-party savers, adapted for SQL Server:

- SQLAlchemy + `pyodbc` transport (sync), same connection-string conventions as `SQLServer_VectorStore` including Entra ID authentication.
- `MERGE INTO ... WHEN MATCHED THEN UPDATE ... WHEN NOT MATCHED THEN INSERT` for the `put` upsert.
- `IF NOT EXISTS (...) INSERT` for the `put_writes` insert-or-ignore path used for non-`WRITES_IDX_MAP` channels.
- `TOP N` instead of `LIMIT N` for `list(..., limit=...)`.
- `JSON_VALUE(metadata, '$.key')` for the metadata `filter=` argument.
- An internal `threading.Lock` guards every session so concurrent graph runs in the same process cannot interleave statements on the underlying engine — mirrors `SqliteSaver`'s safety semantics.

Closes #15

## Public API

```python
from langchain_sqlserver import SQLServerSaver

saver = SQLServerSaver(
    connection_string="Driver={ODBC Driver 18 for SQL Server};Server=tcp:host,1433;Database=mydb;Uid=user;Pwd=pwd;TrustServerCertificate=yes;",
)
graph = builder.compile(checkpointer=saver)
config = {"configurable": {"thread_id": "user-1"}}
graph.invoke({...}, config)
```

Implements `get_tuple` / `list` / `put` / `put_writes` / `delete_thread`, plus a lazy `setup()` for table creation. Tables (`checkpoints`, `checkpoint_writes`) are configurable via `checkpoints_table` / `writes_table` and `db_schema`.

## Files

- `langchain_sqlserver/checkpoint.py` — new module with `SQLServerSaver`.
- `langchain_sqlserver/__init__.py` — export the new class.
- `pyproject.toml` + `uv.lock` — add `langgraph-checkpoint>=2.0,<4.0` to project dependencies.
- `tests/unit_tests/test_checkpoint.py` — 8 unit tests covering construction validation, qualified-name building (schema / no-schema / overrides), Entra-ID-vs-uid/pwd routing, the empty-input fast path on `put_writes`, and SQL-injection guard on the metadata filter. No live DB required.
- `tests/unit_tests/test_imports.py` — extend the `__all__` snapshot.
- `tests/integration_tests/test_checkpoint.py` — 6 integration tests covering: `put` + `get_tuple` round-trip, latest-checkpoint lookup, descending `list` with `limit`, `before=` slicing, `put_writes` retrievability through `get_tuple.pending_writes`, and scoped `delete_thread`. Uses the existing `TEST_AZURESQLSERVER_TRUSTED_CONNECTION` / `TEST_PYODBC_CONNECTION_STRING` env vars.

## Test plan

- [x] `ruff check langchain_sqlserver tests` — clean
- [x] `mypy langchain_sqlserver tests` — clean (14 source files)
- [x] `pytest tests/unit_tests/` — 9 passed (8 new + import snapshot), 1 module skipped (pre-existing)
- [ ] `pytest tests/integration_tests/test_checkpoint.py` against a live SQL Server / Azure SQL instance.

## Notes

- Connection-string parsing and Entra-ID token-provider helpers mirror those on `SQLServer_VectorStore` so users get identical auth semantics across vector store, chat history, and checkpoint saver. A follow-up to factor these into a shared internal module is reasonable, but kept out of this PR so it stays additive.
- `poetry.lock` is intentionally left in tree as-is — `uv.lock` is the canonical lockfile updated for the new dependency; happy to refresh `poetry.lock` if maintainers prefer.


0xDevNinja:
Thanks for the thorough review — pushed a commit addressing the real issues here.

Fixed:
- `idx` is now an `INTEGER` column instead of `NVARCHAR(20)`. Since some writes use negative indices, the old text `ORDER BY idx` sorted lexicographically and could reorder pending writes — good catch.
- `metadata` is now stored as `NVARCHAR(MAX)` JSON text rather than `LargeBinary`. The filter path uses `JSON_VALUE`, and casting binary UTF-8 back to NVARCHAR never reliably round-trips, so metadata filters were effectively broken. Updated the read/write/filter paths to match (dropped the CAST and the utf-8 encode step).
- `put_writes` now defaults `checkpoint_ns` to `""` like `get_tuple`/`list`, so omitting it no longer raises `KeyError`.
- `list()` only emits `TOP n` for a positive limit now, to avoid invalid T-SQL on a negative value.
- Removed the unused `jsonplus_serde` attribute.

Added unit tests for the column types, the `checkpoint_ns` default, and the `TOP` guard.

A few of the other suggestions I left as-is on purpose:
- The Entra ID token hook is already registered per-engine (`event.listen(engine, ...)`), not globally on the `Engine` class with `once=True`, so pooled/reconnecting connections do get a fresh token.
- `_qualified()` already escapes `]` -> `]]` before interpolating, so the identifier quoting is safe.
- The batch-level `replace = all(channel in WRITES_IDX_MAP ...)` decision in `put_writes` intentionally mirrors the upstream `SqliteSaver`/`PostgresSaver` behavior, so I kept it consistent with the first-party savers rather than switching to a per-write decision.

PR #722, opened 2026-06-19
Consolidate Python tooling on uv (remove Poetry)

jkennedyvz:
## Why

The `libs/*` packages were **dual-locked** — both `poetry.lock` and `uv.lock` committed — and built with **Poetry** in CI, while Dependabot was configured for the **uv** ecosystem. Dependabot only ever updated `uv.lock`, so ~105 security alerts pinned to the `poetry.lock` files could never be auto-resolved. This removes Poetry entirely and standardizes on uv, which clears those alerts and gives Dependabot a manifest it actually manages.

Supersedes #721 (the dependabot-config-only fix); the dependabot changes are folded in here.

## Changes

**pyproject (all 6 libs)**
- `build-backend` → `uv_build` (`module-root = ""` for the flat-layout libs; `src` layout for azure-cosmosdb / azure-postgresql).
- Removed every `[tool.poetry*]` table — `[dependency-groups]` + `[tool.uv]` are now the single source of truth. Added them to **azure-cosmosdb**, which had no uv config at all.
- Removed an erroneous empty `libs/azure-cosmosdb/src/__init__.py` (uv_build rejects it as a namespace-parent file; Poetry silently ignored it).
- Refreshed `uv.lock`; deleted all `libs/*/poetry.lock`.

**Makefiles**
- `poetry run` / `poetry sync` → self-contained `uv run --frozen --all-extras --group <...>` targets. `--frozen` keeps the lockfile authoritative so the "tests created no new files" CI check stays green. azure-postgresql keeps its tox-uv flow via `uv sync` + `uv run tox`.

**CI workflows + action**
- `_test`, `_lint`, `_compile_integration_test`: `astral-sh/setup-uv` + uv-driven make/`uv run`; mypy caches and the lock check keyed on `uv.lock`.
- `_release`, `_test_release`: `poetry build` → `uv build`, `poetry version` → `uv version`, `poetry install` → `uv sync`. **The `pypa/gh-action-pypi-publish` trusted-publish step is unchanged.**
- Deleted the now-unused `.github/actions/poetry_setup` action and the vestigial `POETRY_VERSION` env.

**Dependabot**
- azure-ai `pip` → `uv`; added `uv` entries for `/libs/azure-cosmosdb` and `/samples`.

## Local validation (uv 0.11.2)

- `uv build` + `uv lock` for all 6 libs ✅
- Unit tests: azure-ai **983 passed**, azure-cosmosdb **246 passed** ✅
- compile-integration (`pytest -m compile`) ✅
- lint (`ruff` + `mypy`) on azure-storage ✅
- azure-postgresql tox-uv env builds and runs (`ruff check src` clean; pre-existing lint debt in `tests/` is unchanged and still swallowed by the existing `|| true`)

## Reviewer notes / risk

- **Release workflows** were ported command-for-command but only run on a tag/`workflow_dispatch`, so they aren't exercised by this PR's CI. The build→publish path is the important one and `uv build` artifacts were verified locally (wheels contain exactly the expected top-level package). One nuance to watch on the **first real release**: in `pre-release-checks`, `make tests` runs via `uv run --frozen` (syncs the local package from the lock), whereas the old Poetry flow ran against the test-PyPI-installed wheel. The published-package *import* smoke-test still runs against the downloaded wheel.
- Out of scope (still on pip): `samples/react-agent-docintelligence` (poetry.lock) and `samples/rag-storage-document-loaders` (requirements.txt).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Follow-on fixes (surfaced once CI actually ran uv)

- **Reformatted 2 test files** (`azure-dynamic-sessions`, `sqlserver`) — the uv lockfiles pin ruff 0.15 while old Poetry CI used the pinned ruff ~0.5.0; `ruff check` passes everywhere, only the newer formatter reflowed two files.
- **Removed a dead `langchain` tox extra in azure-postgresql** — its tox config referenced `extras = ["langchain"]` in three envs, but that extra was never defined and nothing imports external `langchain`/`langgraph` (only `langchain_core`, already a dep). This had been silently blocking the postgresql tox `make test` at install time; removing it lets the test envs run. (Note: the `type`/mypy tox env has pre-existing type errors, but CI does not run it — `make test`=`tox -m test`, `make lint`=`tox -e lint`.)


marlenezw:
Hi @jkennedyvz, this looks good and build is passing. I just want to run a few more checks on my end and then will merge.

kyleknap (approved):
For the `langchain-azure-storage` package, the changes looked good :ship:; I did not review any of the other packages though. I built the `langchain-azure-storage` sdist and wheels using this new backend and those looked correct and are installable as well.

marlenezw (approved):

PR #801, opened 2026-07-03
feat(sqlserver): use binary collation for custom_id column

### LINKED ISSUE #74: [langchain-sqlserver] Use binary collation for custom_id column
To get the best performance possible, binary collation should be used for the `custom_id` column. Specifically the collation `Latin1_General_100_BIN2_UTF8` should be used when creating the `custom_id` column in the `EmbeddingStore` function.

To keep backward compatibility and allow users to use the default collation defined for the database, instead of forcing the binary collation, a new option should be introduced in the `SQLServer_VectorStore` class so that one can opt-out of the forced binary collation, for example:

```python
MySqlServerVectorStore.ForceBinaryComparisonOnCustomId = false
```

0xDevNinja:
Implements #74.

## What
The `custom_id` column on the vector store table is now created with the binary collation `Latin1_General_100_BIN2_UTF8`, which avoids linguistic collation rules and gives the best comparison/lookup performance (custom ids are opaque identifiers, so linguistic ordering is never needed).

A new `use_binary_collation` flag (default `False` to keep the database's default collation) is added to `SQLServerVectorStore.__init__`. Set it to `True` to use the binary collation instead.

```python
store = SQLServerVectorStore(
    connection_string=...,
    embedding_function=...,
    embedding_length=1536,
    use_binary_collation=False,  # opt out, use DB default collation
)
```

## Notes
- Only affects tables newly created by the store; existing tables are not altered.
- Value `True` preserves the performance intent from the issue while the opt-out keeps backward compatibility for anyone relying on the database default collation.

## Tests
Added `tests/unit_tests/test_binary_collation.py` asserting the `custom_id` column collation matches the flag (None/default when use_binary_collation is False, and a binary collation otherwise). Runs without a live DB.

Closes #74

### DIFF
--- a/libs/sqlserver/langchain_sqlserver/vectorstores.py
+++ b/libs/sqlserver/langchain_sqlserver/vectorstores.py
@@ -149,7 +149,8 @@ class VectorType(UserDefinedType):
 AZURE_TOKEN_URL = "https://database.windows.net/.default"  # Token URL for Azure DBs.
 # Binary collation for the `custom_id` column. A binary (BIN2) collation gives
 # the best lookup/comparison performance since it avoids linguistic collation
-# rules. Applied by default; can be disabled via `force_binary_collation=False`.
+# rules. Opt in via `use_binary_collation=True`; otherwise the database's
+# default collation is used.
 CUSTOM_ID_COLLATION = "Latin1_General_100_BIN2_UTF8"
 DISTANCE = "distance"
 DEFAULT_DISTANCE_STRATEGY = DistanceStrategy.COSINE
@@ -199,7 +200,7 @@ class SQLServerVectorStore(VectorStore):
         relevance_score_fn: Optional[Callable[[float], float]] = None,
         table_name: str = DEFAULT_TABLE_NAME,
         batch_size: int = DEFAULT_BATCH_SIZE,
-        force_binary_collation: bool = True,
+        use_binary_collation: bool = False,
     ) -> None:
         """Initialize the SQL Server vector store.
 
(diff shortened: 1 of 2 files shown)


libs/sqlserver/langchain_sqlserver/vectorstores.py
```diff
@@ -195,6 +199,7 @@ def __init__(
         relevance_score_fn: Optional[Callable[[float], float]] = None,
         table_name: str = DEFAULT_TABLE_NAME,
         batch_size: int = DEFAULT_BATCH_SIZE,
+        force_binary_collation: bool = True,
```
beccadaniel: >To keep backward compatibility and allow users to use the default collation defined for the database, instead of forcing the binary collation, a new option should be introduced in the SQLServer_VectorStore class so that one can opt-out of the forced binary collation

@0xDevNinja , based on the work item this PR addresses, I'd suggest having this as an option instead of a default `True`.

beccadaniel: nit: can we rename this parameter - e.g. `use_binary_collation`?

0xDevNinja:
Good call — better to keep the current behavior for anyone already relying on the default collation. I've renamed the parameter to `use_binary_collation` and flipped the default to `False`, so the binary collation is now opt-in and the `custom_id` column keeps the database's default collation otherwise. Updated the docstring, the inline comments and the unit tests to match. Pushed in 2bc0640.

beccadaniel (approved):

beccadaniel:
@0xDevNinja, the changes look good to me, and I have approved. But before the merge, can you go ahead and update the PR description to reflect this last change?

Never mind, updated.

PR #812, opened 2026-07-06
fix(sqlserver): use not_in for $nin metadata filter operator

0xDevNinja:
## Problem
The `$nin` metadata filter operator is broken. In `_handle_field_filter`, the `$nin` branch calls `queried_field.nin_(...)`:

```python
elif operator in {"$nin"}:
    return queried_field.nin_([str(val) for val in filter_value])
```

SQLAlchemy column elements have no `nin_` method (the correct method is `not_in()`), so **any** similarity search using a `$nin` metadata filter raises at query time:

```
AttributeError: Neither 'Function' object nor 'Comparator' object has an attribute 'nin_'
```

Example that triggers it:
```python
store.similarity_search("q", filter={"color": {"$nin": ["red", "blue"]}})
```

## Fix
Use the correct SQLAlchemy method:
```python
return queried_field.not_in([str(val) for val in filter_value])
```
This compiles to `JSON_VALUE(...) NOT IN (red, blue)` as expected.

## Tests
Added `tests/unit_tests/test_filters.py` covering the `$nin`, `$in`, and `$like` clauses. They compile the filter expression to SQL, so they run without a live database. Ruff + mypy clean; full unit suite passes.

### DIFF
--- a/libs/sqlserver/langchain_sqlserver/vectorstores.py
+++ b/libs/sqlserver/langchain_sqlserver/vectorstores.py
@@ -1213,7 +1213,7 @@ class SQLServerVectorStore(VectorStore):
             if operator in {"$in"}:
                 return queried_field.in_([str(val) for val in filter_value])
             elif operator in {"$nin"}:
-                return queried_field.nin_([str(val) for val in filter_value])
+                return queried_field.not_in([str(val) for val in filter_value])
             elif operator in {"$like"}:
                 return queried_field.like(str(filter_value))
             else:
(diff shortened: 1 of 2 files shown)


beccadaniel (approved):

beccadaniel:
@0xDevNinja , I had to revert this change because it's failing CI - due to the previous `use_binary_collation` work. The failure is in the test added. Would you mind updating this and re-opening the PR?

0xDevNinja:
Sorry about that, and thanks for the heads-up. The test helper builds the store with `__new__`, so it has to set everything `_get_embedding_store` reads — `_use_binary_collation` landed in between my PR being opened and merged, and I missed it.

Re-opened as #853, rebased on current `main` with the helper setting `_use_binary_collation`. Verified the full sqlserver unit suite and `make lint` against post-collation `main` this time.

beccadaniel:
No worries at all. Thanks for addressing it.

PR #813, opened 2026-07-06
fix(sqlserver): preserve MMR ranking order in results

0xDevNinja:
## Problem
`max_marginal_relevance_search_by_vector` returns documents in the store's original fetch order rather than the order chosen by the MMR algorithm.

`maximal_marginal_relevance` returns a list of indices in ranked (relevance/diversity) order, but the result was built as:

```python
return [
    value for idx, value in enumerate(results_as_docs) if idx in mmr_selects
]
```

Iterating `results_as_docs` and filtering by membership re-sorts the selected docs back into ascending fetch order, discarding the MMR ranking.

Example: query `[1, 0]` over docs A`[0.9, 0.1]`, B`[0.0, 1.0]`, C`[1.0, 0.0]` with `k=2`. MMR selects `[2, 0]` → should return `[C, A]`, but the old code returned `[A, C]`.

## Fix
Index into `results_as_docs` by the selected indices so the ranking is preserved:

```python
return [
    results_as_docs[idx] for idx in mmr_selects if idx < len(results_as_docs)
]
```

## Tests
Added `tests/unit_tests/test_mmr.py`, which stubs `_search_store` and asserts the MMR ordering (`[C, A]`). Runs without a live database. Ruff + mypy clean; unit suite passes.


0xDevNinja:
Both good catches, thanks — fixed in a5af8a9.

1. Bounds check now requires `0 <= idx < len(results_as_docs)`. You're right that a -1 would have wrapped to the last element and quietly masked an upstream MMR failure.
2. Confirmed the tie: with query `[1, 0]`, the first pick is `C`, and at `lambda_mult=0.5` both remaining candidates score exactly 0 (relevance and redundancy cancel — `A` is 0.5*0.994 - 0.5*0.994, `B` is 0.5*0 - 0.5*0), so the order was only holding by tie-break. Switched the test to `lambda_mult=0.7`, where `A` scores ~0.398 vs `B` at 0, so `[C, A]` follows from the ranking itself.

Also merged current `main` into the branch so this is verified against the `use_binary_collation` changes.

PR #817, opened 2026-07-08
fix(azure-ai): return empty list from async MMR search when no results

0xDevNinja:
### Description

`AzureSearch.amax_marginal_relevance_search` / `amax_marginal_relevance_search_with_score` crash with `ValueError` when the underlying async search returns **zero** candidates (empty index, or a filter that matches nothing).

The async reranking helper builds the candidate list and then immediately unpacks it:

```python
docs = [(...) async for result in results]
documents, scores, vectors = map(list, zip(*docs))   # <-- zip(*[]) -> unpack 0 into 3
```

With no results, `zip(*[])` yields nothing, so the three-way unpack raises:

```
ValueError: not enough values to unpack (expected 3, got 0)
```

The **synchronous** twin `_reorder_results_with_maximal_marginal_relevance` already handles this with an early `if not docs: return []`; the async version was missing that guard, so the two paths diverge for the empty case.

Repro:

```
sync empty  -> []
async empty -> RAISED ValueError not enough values to unpack (expected 3, got 0)
```

### Fix

Add the same `if not docs: return []` guard to `_areorder_results_with_maximal_marginal_relevance`, so the async MMR search returns an empty list — matching the sync path — instead of raising.

### Testing

Added two unit tests to `tests/unit_tests/test_azure_search.py` (guarded by the existing `@pytest.mark.requires("azure.search.documents")` marker): one asserts the sync helper returns `[]` for empty input (regression parity), and one drives the async helper with an empty async generator and asserts `[]`. The async test uses `asyncio.run`, so no new test dependency is introduced. `ruff`, `ruff format`, and `mypy` are clean.

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/vectorstores/azuresearch.py
+++ b/libs/azure-ai/langchain_azure_ai/vectorstores/azuresearch.py
@@ -2020,6 +2020,8 @@ async def _areorder_results_with_maximal_marginal_relevance(
         )
         async for result in results
     ]
+    if not docs:
+        return []
     documents, scores, vectors = map(list, zip(*docs))
 
     # Get the new order of results.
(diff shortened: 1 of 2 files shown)


santiagxf:
@farzad528 can you take a look at this issue?

PR #840, opened 2026-07-15
feat(cosmosdb): fast-path delta channel history for LangGraph checkpointer

gwincr11:
Override get_delta_channel_history / aget_delta_channel_history to replace the base per-ancestor get_tuple walk (O(N) serial round trips) with two bulk queries, walking the parent chain in memory. Consolidate the shared "parse writes keys -> sort by idx -> load" logic into a single _load_sorted_writes helper reused by _group_pending_writes and both _load_pending_writes methods.

Bump langgraph-checkpoint to >=4.1.0,<5.0.0 for DeltaChannelHistory and add sync/async parity unit tests plus integration tests.

https://www.langchain.com/blog/delta-channels-evolving-agent-runtime


libs/azure-cosmosdb/src/langchain_azure_cosmosdb/_langgraph_checkpoint_store.py
```diff
@@ -570,6 +747,72 @@ def list(
...
+        ancestors = _order_on_path_ancestors(docs_by_checkpoint, target_checkpoint_id)
+        if not ancestors:
+            return _reconstruct_delta_channel_history(
+                self.cosmos_serde, channels, [], {}
+            )
+
+        writes_prefix = _make_writes_partition_prefix(thread_id, checkpoint_ns)
+        write_docs = list(
+            self.container.query_items(
+                query=("SELECT * FROM c WHERE STARTSWITH(c.partition_key, @prefix)"),
```
aayush3011, line 806: **🟡 Recommendation** · Performance: Scalability

**Delta fast-path fetches & deserializes the entire thread's writes, not just on-path ancestors**

This query pulls **all** writes for the thread/ns, and `_group_pending_writes` (line 810) then fully decodes every one of them via `_load_sorted_writes`:

```python
write_docs = list(self.container.query_items(
    query=("SELECT * FROM c WHERE STARTSWITH(c.partition_key, @prefix)"),
    parameters=[{"name": "@prefix", "value": writes_prefix}],
))
writes_by_checkpoint = _group_pending_writes(self.cosmos_serde, write_docs)
```

However, `_reconstruct_delta_channel_history` only consumes writes for the on-path ancestors and stops once every channel is seeded. The base `get_delta_channel_history` you're overriding loads writes **only for the ancestors it actually visits** and terminates early once `remaining` is empty.

Net effect: round-trips correctly drop from O(N) to 2 (the intended win), but deserialization + RU now scale with **total thread history** (every fork, full depth) instead of walk depth. For a long-lived or branchy thread whose seed sits at the immediate parent, this can decode dramatically more than the base did — on the exact hot path this PR optimizes.

**Suggestion:** once `ancestors` is known, scope the writes query to just those checkpoints (e.g. an `IN` / `ARRAY_CONTAINS` over the on-path write partition-key prefixes) or decode lazily per consumed checkpoint, so cost stays proportional to walk depth. The same applies to the async path (`aio/_langgraph_checkpoint_store.py:326`).

---
<sub>⚠️ AI-assisted review — may be incorrect. Agree? → resolve the conversation. Disagree? → reply with your reasoning.</sub>

libs/azure-cosmosdb/tests/unit_tests/test_langgraph_checkpoint_delta.py
```diff
@@ -0,0 +1,137 @@
...
+
+def _build_saver() -> CosmosDBSaverSync:
+    saver = object.__new__(CosmosDBSaverSync)
+    BaseCheckpointSaver.__init__(saver)
+    saver.container = FakeSyncContainer()
+    saver.cosmos_serde = make_serde()
+    return saver
+
+
+def _seed_chain(saver: CosmosDBSaverSync, depth: int) -> str:
```
aayush3011, line 37: **🟢 Suggestion** · Testing: Missing Scenario (Forked Threads)

**Parity tests only seed linear chains — add a forked-thread case to prove off-path exclusion**

The `override == base` assertions are spot-on, and I can see latest-resolution (`test_override_matches_base_latest_checkpoint`), empty channels, and missing-thread are all covered — nice. The remaining gap: `_seed_chain` only builds a single **linear** parent chain, so **forked threads are never exercised**.

That case matters specifically because of the writes query in `_langgraph_checkpoint_store.py:806`, which intentionally fetches writes for **all** checkpoints in the thread — including off-path branches. Correctness then hinges entirely on `_reconstruct_delta_channel_history` walking only the on-path parent chain and ignoring the siblings. A linear-chain test cannot catch a regression there.

**Suggestion:** add a parity test that builds a fork — two children of a common parent, each with their own writes — queries one child, and asserts `override == base`, proving the sibling branch's writes don't leak into the result.

---
<sub>⚠️ AI-assisted review — may be incorrect. Agree? → resolve the conversation. Disagree? → reply with your reasoning.</sub>

aayush3011 (changes requested):
Thanks @gwincr11 — this is a well-constructed fast-path, and calling the base implementation directly to assert `override == base` is exactly the right way to prove equivalence. I verified the core walk faithfully matches `BaseCheckpointSaver.get_delta_channel_history`, the partition scheme is correct, and the async path safely reuses the sync helpers — nice work.

I'm **requesting changes** on two items before merge:

1. 🟡 **Scalability** — the writes fast-path fetches and deserializes the *entire* thread's writes (all branches, full history), even when only a few on-path ancestors are consumed. Inline on `_langgraph_checkpoint_store.py`.
2. 🟢 **Test gap** — the parity tests only seed linear chains; a forked-thread test is needed to prove off-path branches are excluded. Inline on the unit test.

Details inline — happy to discuss.

---
<sub>⚠️ AI-assisted review — may be incorrect. Agree? → resolve the conversation. Disagree? → reply with your reasoning.</sub>

gwincr11:
@aayush3011 just a heads up I pushed the requested changes. Thanks!

PR #878, opened 2026-07-30
feat(azure-storage): conform AzureBlobBackend to deepagents 0.7.0

darieldatoon:
Updates `AzureBlobBackend` for the `BackendProtocol` changes in [deepagents 0.7.0](https://www.langchain.com/blog/deep-agents-v0-7). `write` now replaces existing files, `delete` is supported, and `grep` accepts a `max_count` cap.

---

## Why now

Our 1.1.0 release pins `deepagents>=0.6.12,<1`. We probably should have set an upper limit of `<0.7` so that users don't get auto-upgraded to 0.7.0.

## What changed

**`write`/`awrite` now overwrite an existing file.**
0.7.0 removed the create-only contract and shipped no compatibility mode. Its `write_file` tool description tells the model the file will be replaced. We were still returning "already exists", which the agent had no way to act on.

**Added `delete`/`adelete`.**
It removes the path plus everything nested under it, and returns not-found for a missing path. Same semantics as `StateBackend` and `StoreBackend`.

**`grep`/`agrep` accept `max_count` and set `truncated`.**
This clears two mypy override errors against the new protocol signature.

It also cuts egress. When a cap is set we scan candidates in chunks and stop once we pass it, rather than downloading every blob in the container and letting the middleware throw away all but 1000 matches. Uncapped scans dispatch everything at once, since there's nothing to stop for. Concurrency is bounded either way, by the thread pool on the sync side and a semaphore on the async side.

`max_count` is clamped to zero or higher. A negative value would be read as a slice from the end, dropping a match while still reporting success.

**Guarded `read(..., limit=0)`.**
This appears to be a regression. On 0.7.0 this raises `ValueError` for any non-empty file while previously it returned an empty read. Filed upstream as [deepagents#5180](https://github.com/langchain-ai/deepagents/issues/5180). The code comment links the issue so the workaround can be removed later.

**Raised two pins.**
`deepagents>=0.7.0,<0.8`, matching what quickjs, daytona, modal, vercel-sandbox, and runloop all moved to. `wcmatch>=11.0`, matching deepagents' own floor, which is why we pinned it in the first place.

**Read pagination now works.**
`slice_read_response` returns `total_lines`, `start_line`, `end_line`, and `next_offset` in 0.7.0. We pick those up with no code change, so `read_file` can tell the model how many lines remain.

## What users will notice

`AzureBlobBackend` is `@beta` and 1.1.0 is a day old, so the exposure is small. Still worth calling out:

- Anyone calling `write()` directly loses the create-guard. The change is silent, with no error.
- Agents now get a recursive delete tool pointed at a live container. `delete("/")` with no `prefix` set empties the whole thing. The class docstring and README cover the mitigations: set a `prefix`, turn on soft-delete or versioning, or leave `delete` out of `FilesystemMiddleware(tools=...)`.
- The pin narrows at both ends. Anyone on deepagents 0.6.x now has to take 0.7.0, which brings its own breaks: no default `TodoListMiddleware`, lean default prompts, and `virtual_mode=True`.
- Python 3.10 users and document-loader users are unaffected.

## Sample fix

`composite_with_memories.py` seeded `AGENTS.md` by writing unconditionally and ignoring the file-exists error. Under the new write semantics that wipes accumulated memory on every run, so it now seeds only when the file is absent.

## Testing

229 unit tests and 99 integration tests against Azurite. `ruff check` and `mypy` are clean on both the package and the tests.

### DIFF
--- a/libs/azure-storage/README.md
+++ b/libs/azure-storage/README.md
@@ -252,7 +252,7 @@ Recommended mitigations:
 
 - Set a `prefix` so the agent can only reach its own key namespace, never the whole container.
 - Enable [soft delete](https://learn.microsoft.com/azure/storage/blobs/soft-delete-blob-overview) and/or [blob versioning](https://learn.microsoft.com/azure/storage/blobs/versioning-overview) on the container so destructive tool calls are recoverable.
-- Scope the agent's credential to the container it should be able to modify.
+- Scope the agent's credential to the container it should be able to modify, following the [Azure RBAC best practices](https://learn.microsoft.com/azure/role-based-access-control/best-practices): assign the role at the container scope rather than the subscription or storage account, and pick the least-privileged built-in role for what the agent actually does — `Storage Blob Data Reader` for a read-only agent, `Storage Blob Data Contributor` only when it must write or delete.
 - Drop or gate the tools you don't want. Omit `delete` from the filesystem middleware, or add a Deep Agents permission rule that denies it or routes it through a human-approval interrupt:
 
   ```python
--- a/libs/azure-storage/README.md
+++ b/libs/azure-storage/README.md
@@ -196,7 +196,7 @@ container_loader = AzureBlobStorageLoader(
 
 ## Deep Agents Azure Blob Storage Backend Usage
 
-[Deep Agents](https://github.com/langchain-ai/deepagents) exposes a `BackendProtocol` — a pluggable interface for file operations (`read`, `write`, `edit`, `ls`, `glob`, `grep`, plus batch upload/download) that an agent uses as its virtual filesystem. This package provides `AzureBlobBackend`, an Azure Blob Storage implementation of that interface, so a deep agent can persist its workspace in a blob container.
+[Deep Agents](https://github.com/langchain-ai/deepagents) exposes a `BackendProtocol` — a pluggable interface for file operations (`read`, `write`, `edit`, `delete`, `ls`, `glob`, `grep`, plus batch upload/download) that an agent uses as its virtual filesystem. This package provides `AzureBlobBackend`, an Azure Blob Storage implementation of that interface, so a deep agent can persist its workspace in a blob container.
 
 The backend requires the optional `deepagents` extra (which itself requires Python 3.11+):
 
@@ -239,7 +239,36 @@ Runnable examples — including a workspace that persists across agent lifetimes
 composite agent with memory and subagents — live in
 [`samples/deepagents-storage-backend/`](../../samples/deepagents-storage-backend/README.md).
 
-File content is stored as UTF-8 text in blob bodies (binary uploads are preserved as bytes). Directories are synthesized from blob key prefixes (no directory marker blobs are created). The backend exposes both synchronous methods (`read`, `write`, `edit`, `ls`, `glob`, `grep`, `upload_files`, `download_files`) and their `a`-prefixed async counterparts (`aread`, `awrite`, …).
+File content is stored as UTF-8 text in blob bodies (binary uploads are preserved as bytes). Directories are synthesized from blob key prefixes (no directory marker blobs are created). The backend exposes both synchronous methods (`read`, `write`, `edit`, `delete`, `ls`, `glob`, `grep`, `upload_files`, `download_files`) and their `a`-prefixed async counterparts (`aread`, `awrite`, …).
+
+### Writes and deletes are destructive
+
+Two backend operations replace or remove data that is already in your container. Both follow the Deep Agents `BackendProtocol` contract, and both are driven by the agent:
+
+- **`write` replaces an existing file in full.** There is no create-only mode — `write_file` on a path that already exists overwrites it rather than erroring. Use `edit` when existing content must be preserved.
+- **`delete` is recursive.** Deleting a directory removes it and everything nested under it. Deleting `"/"` removes every blob in the configured `prefix` namespace, or — if no `prefix` is set — **every blob in the container**.
+
+Recommended mitigations:
+
+- Set a `prefix` so the agent can only reach its own key namespace, never the whole container.
+- Enable [soft delete](https://learn.microsoft.com/azure/storage/blobs/soft-delete-blob-overview) and/or [blob versioning](https://learn.microsoft.com/azure/storage/blobs/versioning-overview) on the container so destructive tool calls are recoverable.
+- Scope the agent's credential to the container it should be able to modify.
+- Drop or gate the tools you don't want. Omit `delete` from the filesystem middleware, or add a Deep Agents permission rule that denies it or routes it through a human-approval interrupt:
+
+  ```python
+  from deepagents import FilesystemMiddleware, create_deep_agent
+
+  agent = create_deep_agent(
+      backend=backend,
+      middleware=[
+          # Register every filesystem tool except `delete`.
+          FilesystemMiddleware(
+              backend=backend,
+              tools=["ls", "read_file", "write_file", "edit_file", "glob", "grep"],
+          )
+      ],
+  )
+  ```
 
 ### Authentication
 
--- a/libs/azure-storage/langchain_azure_storage/deepagents/backend.py
+++ b/libs/azure-storage/langchain_azure_storage/deepagents/backend.py
@@ -203,12 +203,8 @@ def _read_result_from_bytes(
 
     Returns raw (unformatted) content -- the Deep Agents middleware applies
     line numbering, the empty-file reminder, and base64 multimodal handling at
-    the tool boundary. For text content, ``slice_read_response`` also supplies
-    the pagination metadata (``total_lines``/``start_line``/``end_line``/
-    ``next_offset``), clamps degenerate windows a model can ask for (negative
-    ``offset`` reads from line 1, non-positive ``limit`` yields an empty read
-    flagged ``no_lines_requested``), and the middleware reports all of that
-    back to the model.
+    the tool boundary. ``slice_read_response`` supplies the pagination metadata
+    and clamps degenerate windows.
 
     Encoding is chosen the way the filesystem-like reference backends do it:
     files whose extension classifies as non-text (image/audio/video/file) are
@@ -268,15 +264,12 @@ def _delete_keys(
 ) -> list[str]:
     """Select the blob keys a recursive delete of one path should remove.
 
-    The listing that produced *blobs* is prefix-narrowed server-side for
-    efficiency, but a bare prefix also matches siblings that merely share a
-    name stem. This filter is what enforces the subtree boundary: a key
-    qualifies only by being the path's own blob or by sitting under its
-    slash-terminated prefix. See `AzureBlobBackend._delete_scope`.
+    The server-side listing prefix also matches siblings sharing a name stem
+    (``/src`` returns ``/src-backup``), so this filter is what enforces the
+    subtree boundary. Do not widen it to a bare prefix match.
 
-    Unlike `ls`, pseudo-directory marker blobs (keys ending in ``/``) are not
-    skipped: they are real blobs inside the subtree being removed, so a
-    recursive delete has to take them with it.
+    Marker blobs (keys ending in ``/``) are deliberately kept, unlike in `ls`:
+    they sit inside the subtree being removed.
     """
     return [
         blob.name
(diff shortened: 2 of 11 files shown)


libs/azure-storage/langchain_azure_storage/deepagents/backend.py
```diff
@@ -260,6 +277,80 @@ def _grep_candidates(
     return candidates
 
 
+def _delete_keys(
```
anjaliratnam-msft: Maybe this function should be renamed to something like `_keys_to_delete` or something similar since it's not actually deleting anything.

samples/deepagents-storage-backend/composite_with_memories.py
```diff
@@ -67,25 +67,28 @@
...
     Written through the composite backend, so ``MEMORY_PATH`` is the same path
     the agents see. The ``/memories/`` route sends it to Azure Blob Storage.
 
     The agents can rewrite this file themselves — memory is left writable so the
-    workspace can accumulate conventions across sessions.
-
-    ``awrite`` fails if the file already exists; that's fine here, since an
-    existing AGENTS.md is sufficient for this example.
+    workspace can accumulate conventions across sessions. Seeding is therefore
+    guarded by a read: since deepagents 0.7.0, ``awrite`` replaces an existing
```
kyleknap, line 77: I think it is fine if we just skip the `aread()` call here and just always write to the file given overrides are by default and this will just help make the sample simpler.

libs/azure-storage/README.md
```diff
@@ -239,7 +239,36 @@ Runnable examples — including a workspace that persists across agent lifetimes
...
+  ```python
+  from deepagents import FilesystemMiddleware, create_deep_agent
+
+  agent = create_deep_agent(
+      backend=backend,
+      middleware=[
+          # Register every filesystem tool except `delete`.
+          FilesystemMiddleware(
+              backend=backend,
+              tools=["ls", "read_file", "write_file", "edit_file", "glob", "grep"],
```
kyleknap, line 267: Would it be possible to defer supporting `delete()` for now to unblock 0.7.0 support? Based on the docs,  `delete()` appears to be an optional method to implement and not needed to support 0.7.0?

Mainly I want to think a bit more how we want to support `delete()`, I'm wary that if we automatically offer `delete()` by default it could be really problematic if someone working with `deepagents` is not aware about it and points the agent at a container and accidentally deletes a bunch of blobs. I'm even thinking if we could explore making it an opt-in constructor parameter to be able to turn on deletes to be a bit more defensive.

Would it work if we put the `delete()` implementation in a separate branch/PR to consider later? I'd also like to see some customer asks potentially for delete before adding especially since this backend is more aimed for persistence. Thoughts?

libs/azure-storage/README.md
```diff
@@ -239,7 +239,36 @@ Runnable examples — including a workspace that persists across agent lifetimes
 composite agent with memory and subagents — live in
 [`samples/deepagents-storage-backend/`](../../samples/deepagents-storage-backend/README.md).
 
-File content is stored as UTF-8 text in blob bodies (binary uploads are preserved as bytes). Directories are synthesized from blob key prefixes (no directory marker blobs are created). The backend exposes both synchronous methods (`read`, `write`, `edit`, `ls`, `glob`, `grep`, `upload_files`, `download_files`) and their `a`-prefixed async counterparts (`aread`, `awrite`, …).
+File content is stored as UTF-8 text in blob bodies (binary uploads are preserved as bytes). Directories are synthesized from blob key prefixes (no directory marker blobs are created). The backend exposes both synchronous methods (`read`, `write`, `edit`, `delete`, `ls`, `glob`, `grep`, `upload_files`, `download_files`) and their `a`-prefixed async counterparts (`aread`, `awrite`, …).
+
+### Writes and deletes are destructive
```
kyleknap, line 244: I like this section. I think generalizing it to be a general note on destructive operations would probably make sense as we should probably classify `edit()` as a destructive operation since it will override original content.

libs/azure-storage/README.md
```diff
@@ -239,7 +239,36 @@ Runnable examples — including a workspace that persists across agent lifetimes
...
+
+Two backend operations replace or remove data that is already in your container. Both follow the Deep Agents `BackendProtocol` contract, and both are driven by the agent:
+
+- **`write` replaces an existing file in full.** There is no create-only mode — `write_file` on a path that already exists overwrites it rather than erroring. Use `edit` when existing content must be preserved.
+- **`delete` is recursive.** Deleting a directory removes it and everything nested under it. Deleting `"/"` removes every blob in the configured `prefix` namespace, or — if no `prefix` is set — **every blob in the container**.
+
+Recommended mitigations:
+
+- Set a `prefix` so the agent can only reach its own key namespace, never the whole container.
+- Enable [soft delete](https://learn.microsoft.com/azure/storage/blobs/soft-delete-blob-overview) and/or [blob versioning](https://learn.microsoft.com/azure/storage/blobs/versioning-overview) on the container so destructive tool calls are recoverable.
```
kyleknap, line 254: I'd probably recommend highlighting blob versioning more if we decide to pass on deletes as part of this PR.

libs/azure-storage/README.md
```diff
@@ -239,7 +239,36 @@ Runnable examples — including a workspace that persists across agent lifetimes
...
+- Scope the agent's credential to the container it should be able to modify.
```
kyleknap: For this one, we could probably just link to the best practices for scoping permissions: https://learn.microsoft.com/en-us/azure/role-based-access-control/best-practices and maybe call out examples of scoping permissions to only resources needed by the agent and using the appropriate role (e.g., `Storage Blob Data Reader` for read only access).

libs/azure-storage/README.md
```diff
@@ -239,7 +239,36 @@ Runnable examples — including a workspace that persists across agent lifetimes
...
+- Drop or gate the tools you don't want. Omit `delete` from the filesystem middleware, or add a Deep Agents permission rule that denies it or routes it through a human-approval interrupt:
```
kyleknap, line 256: Should we also include file system permission objects? https://docs.langchain.com/oss/python/deepagents/permissions this seems to be the more public facing interface from the documentation and probably better if you want a blanket allow/deny list of tools.

libs/azure-storage/langchain_azure_storage/deepagents/backend.py
```diff
@@ -217,12 +220,26 @@ def _read_result_from_bytes(
...
+                # Workaround for a deepagents 0.7.0 bug: for a non-empty file,
+                # `slice_read_response(fd, offset, 0)` builds a zero-length
+                # window (`start_line=offset+1`, `end_line=offset`) that its own
+                # `ReadResult.__post_init__` then rejects with a ValueError. The
+                # `read_file` tool declares `limit` as a plain int with no lower
+                # bound, so a model can reach it. Return the empty window the
+                # caller asked for instead of raising.
+                #
+                # Tracked upstream at
+                # https://github.com/langchain-ai/deepagents/issues/5180.
```
kyleknap: It looks like this is now fixed? https://github.com/langchain-ai/deepagents/issues/5180

I wonder if it makes sense to hold off pulling this workaround until a patch of deepagents comes out and we can make the floor `0.7.1`?

kyleknap (commented):
Looks good. Thanks for putting this together so quickly!

I think the biggest feedback I had was on whether we want to start supporting `delete` as it seems its optional for the protocol. I'd probably prefer it get requested first as a feature request before we add it since it is a fairly dangerous operation in the context of Blob storage especially since it is a recursive delete. Then if we add it, it may make sense to require a flag to enable it in the backend so it is not tuned on by default to help users avoid missing the backend does have that capability. Happy to discuss this more though.

libs/azure-storage/langchain_azure_storage/deepagents/backend.py
darieldatoon: Yup! `0.7.1` was just released today. So I'll remove this workaround and update the floor.

darieldatoon:
> I think the biggest feedback I had was on whether we want to start supporting `delete` as it seems its optional for the protocol.

@kyleknap, I understand the concern around the destructive nature of `delete()`. Here's my reasoning for including it. All the built-in backends implement it (`StateBackend`, `StoreBackend`, `FilesystemBackend`, `ContextHubBackend`). I think it is the backend's job to provide the API/tools, and the harness job to make it safe. You can remove `delete()` from the harness completely, use middleware to enforce file permissions, or add a programmatic hook to enforce HITL on `delete()` tool calls.

But also understand your viewpoint that a developer would have to know to do these things.

Let me know where you land on this one and I'll update the PR accordingly.

anjaliratnam-msft (commented):
This looks good! I just had one small comment that was more of a nitpick.

libs/azure-storage/langchain_azure_storage/deepagents/backend.py
darieldatoon: Updated.

libs/azure-storage/README.md
kyleknap: Talked about this offline, but we are going to go forward with adding `delete()` support. Specifically, given that all of the other backends support delete(), it would likely cause more issues in being inconsistent. Also given the Deep Agents harness has its own safeguards/opinions on top of these tools, it is much more safe than offering `delete()` say as an MCP tool that could be arbitrarily used.

libs/azure-storage/README.md
kyleknap: @darieldatoon I think at least adding this above link would be useful to include for readers to learn more here.

libs/azure-storage/langchain_azure_storage/deepagents/backend.py
```diff
@@ -829,6 +887,127 @@ async def aedit(
...
+        )
+        keys = _keys_to_delete(blobs, exact_key, descendant_prefix)
+        if not keys:
+            return DeleteResult(error=f"Error: File '{file_path}' not found")
+
+        def remove(key: str) -> str | None:
+            try:
+                container.get_blob_client(key).delete_blob()
+            except ResourceNotFoundError:
+                return None  # Raced with another deleter; treat as done.
```
kyleknap, line 926: Are there cases within `deepagents` where it will try to do concurrent deletes on the same file path? Mainly curious if there is a specific scenario within an agent workflow where we'd need to swallow this exception as I'm also concerned that we could possibly swallowing any future bugs in not propagating the resource not found error.

libs/azure-storage/tests/unit_tests/deepagents/backend/test_delete.py
```diff
@@ -0,0 +1,423 @@
...
+        container.list_blobs = async_list(
+            [
+                make_blob("pfx/src", 5),
+                make_blob("pfx/src/a.py", 5),
+                make_blob("pfx/src/deep/b.py", 5),
+            ]
+        )
+        blob = AsyncMock(spec=AsyncBlobClient)
+        container.get_blob_client.return_value = blob
+        result = await backend.adelete("/src")
```
kyleknap, line 79: I think the logic currently handles this, but it may be worth to add a case where a `/` is added to the end of the path to explicitly indicate that it is a prefix that is trying to be deleted (e.g., `adelete("/src/")`) and ideally in this scenario, it would not delete the `pfx/src` blob as well since that is technically not part of that prefix.

kyleknap (commented):
Looks good! I just had a few more comments and I think we should be set here.

libs/azure-storage/README.md
darieldatoon: Added

libs/azure-storage/langchain_azure_storage/deepagents/backend.py
darieldatoon: This could happen if it tries to delete a file and its parent folder concurrently at the same time. This could create a race condition.

Kept as is.

libs/azure-storage/tests/unit_tests/deepagents/backend/test_delete.py
darieldatoon: Actually, the opposite behavior is currently implemented. `_validate_path` delegates to deepagents' `validate_path`, which runs `os.path.normpath`. And that strips the trailing slash.

This is consistent with `StateBackend` and `StoreBackend` which both open `delete` with `base = file_path.rstrip("/")`.

I can understand your point given that Blob Storage is not a native filesystem. So I can understand if we wanted to diverge from other backend's behavior.

Let me know which direction you want to take. For now, added some tests to make the current behavior clear.

kyleknap: Oh interesting, but the `StoreBackend` is just a key-value store right? Blob Storage is probably more aligned with the `StoreBackend` in that they are both effectively key-value stores. So, I'd probably lean more towards staying consistent with the `StoreBackend`. Also given all paths provided to the backend go through `validate_path`, it would be weird if the backend inconsistently resolved itself for `delete()` but not the rest of the backend methods.

kyleknap (approved):
Looks good! :ship: Thanks for again for the fast turnaround in supporting 0.7 and the effort here!

PR #879, opened 2026-07-30
Pin `langchain-azure-storage` `deepagents` dependency ceiling to `0.7.0`

kyleknap:
There were incompatible changes to interfaces introduced in 0.7.0 that need to be accounted for in the Azure Storage Deep Agents backend. The pin prevents automatic upgrades to 0.7.0 while proper support is added.

The PR also bumps the `langchain-azure-storage` version to `1.1.1` to make this update available in a patch release.


darieldatoon (approved):

anjaliratnam-msft (approved):
Looks good!

Issue #906, opened 2026-08-04
Consolidate the Azure Container Apps integrations into langchain-azure-compute


darieldatoon:
# Consolidate the Azure Container Apps integrations into `langchain-azure-compute`

## Summary

Azure Container Apps offers two distinct compute products for running agent-authored code:

| Product | ARM resource | Today |
|---|---|---|
| Dynamic sessions | `Microsoft.App/sessionPools` | `langchain-azure-dynamic-sessions` (tools + a Deep Agents backend) |
| Sandboxes | `Microsoft.App/sandboxGroups` | no integration |

I'd like to consolidate both into a single new package, `langchain-azure-compute` and deprecate `langchain-azure-dynamic-sessions` in favour of it.

> **Name.** This was originally proposed as `langchain-azure-container-apps`. Per [@ShichaoQiu's comment](https://github.com/langchain-ai/langchain-azure/issues/906#issuecomment-5288601146), the package is `langchain-azure-compute` — a name with room for Azure compute products beyond Container Apps as the consolidation continues. The stack below has been renamed accordingly; its scope is unchanged and still covers only the two Container Apps products.

## Why

**Sandboxes have no integration currently.** They are the stateful counterpart to dynamic sessions — suspend/resume, snapshots, persistent volumes, VNet integration — and they are the better fit for a long-running agent workspace.

**The existing dynamic-sessions backend is broken on deepagents 0.7.** `SessionsBashBackend` in `langchain-azure-dynamic-sessions` 1.0.3 targets 0.6:

- it overrides `ls_info`/`glob_info`, which no longer exist in 0.7, so both overrides are silently ignored and the base class's `python3 -c` commands run instead — commands a Shell-typed session pool image may not be able to run;
- `read()` returns a `str` where 0.7 callers expect a `ReadResult`.

## Proposed layout

```
libs/azure-compute/
└── langchain_azure_compute/
    ├── dynamic_sessions/        # [dynamic-sessions] extra
    │   ├── tools/               #   SessionsPythonREPLTool, SessionsBashTool
    │   └── backends/            #   SessionsBashBackend  (Deep Agents)
    └── sandboxes/               # [sandboxes] extra
                                 #   ACASandbox           (Deep Agents)
```

Extras gate *dependencies*, not module presence: every module ships in the wheel and each subpackage checks its own requirements at import time, raising an error that names the extra to install. Same pattern as `langchain-azure-storage`.

## Plan

[...]

PR #907, opened 2026-08-04
feat(compute): scaffold langchain-azure-compute package

darieldatoon:
# Scaffold `langchain-azure-compute`

Part of the stack for #906 · **1 of 6** · base `main`

Consolidates the Azure Container Apps integrations into one package. This PR adds the empty package and its repository wiring; nothing user-facing yet.

## What to review

Mechanical: `check_diff.py` registration, README/CONTRIBUTING entries, `maintainers.md` row, LICENSE, `scripts/check_imports.py`.

Worth a look:

- **`pyproject.toml`** — base dependencies only. Extras arrive alongside the code that needs them, so installing one integration never pulls the other's dependencies.
- **`Makefile`** — `integration_tests` is defined rather than commented out because `_release.yml` invokes it unconditionally. Integration tests added later must *skip* on missing environment variables rather than fail.
- **Coverage gate** — `fail_under = 100` plus a `make coverage` target. Gating from the start forces each later PR in this stack to arrive with the tests for the code it adds, and the gate holds at every one of them. It is a local/review-time check: the shared CI workflows run `make test`, not this target, so nothing in CI enforces the floor. The Makefile says so too.

`make coverage` is deliberately not folded into `make test`: the deepagents-dependent modules are absent on Python 3.10, where their tests skip and coverage would fall below the floor.

## Behaviour changes

None. No integration code.

## Validation

- 1 unit test, 100% statement and branch coverage
- `make check_imports`, `make spell_check`

## Notes

`uv.lock` is generated (+2,094 lines). It appears in full because the package is new; the declaration that produced it is the `dependencies` block in `pyproject.toml`.

The package README and `__init__` docstring describe only what this PR ships. Both grow in the PRs that add the integrations, so no layer of this stack documents a feature that is not yet present.


ShichaoQiu (approved):

ShichaoQiu:
Thanks @darieldatoon , LGTM, I have approved it.

PR #908, opened 2026-08-04
refactor(compute): migrate the dynamic sessions tools

darieldatoon:
# Migrate the dynamic sessions tools

Part of the stack for #906 · **2 of 6** · base: *Scaffold `langchain-azure-compute`*

Brings `SessionsPythonREPLTool` and `SessionsBashTool` into the consolidated package behind a `[dynamic-sessions]` extra.

## Review in this order

| Production | Its tests |
|---|---|
| `dynamic_sessions/_common.py` — user agent, token provider, URL builder | `test_import_guards.py`, URL cases in the tool tests |
| `dynamic_sessions/_base.py` — shared fields + session lifecycle | lifecycle cases in both tool test modules |
| `dynamic_sessions/tools/sessions.py` — the two tools | `test_sessions_bash_tool.py`, `test_sessions_python_repl_tool.py`, `test_sessions_files.py` |
| `dynamic_sessions/__init__.py` — `requests` import guard | `test_imports.py`, `test_import_guards.py` |

## Why the dedup

The two tools were near-clones — identical fields, session-id lock and rotation, sync/background deletion, `delete_session`, `close`, and the context-manager protocol. Those now live once on a private `_BaseSessionsTool`. Each class keeps its own request bodies, response parsing and file-API shapes, which is where they genuinely differ.

`_common.py` sits below both so the backend in the next PR does not have to import a tool module to authenticate — the old package had that inversion.

## Intentional behaviour changes

1. **Endpoints with a query string are now rejected.** One `build_session_url` replaces four hand-rolled copies. The old code claimed to support a query-bearing endpoint but appended the resource path *after* the query: `https://host/pool?x=1/executions&identifier=…`. No real pool management endpoint carries a query or fragment.
2. **The session identifier is fully percent-encoded**, so a value containing `/` can no longer extend the URL path (`a b/c` → `a%20b%2Fc`).
3. **The user agent reports this distribution.** It previously asked metadata for `langchain-azure-dynamic-sessions`; in an environment holding only the new package that resolved to `langchain-azure-dynamic-sessions/0.0.0`.
4. **`upload_file` no longer leaks a descriptor.** The Python tool opened the file and never closed it. Both tools now share one `_upload_payload` helper that closes only a handle it opened and leaves a caller-supplied stream alone. This leak also ships in `langchain-azure-dynamic-sessions` 1.0.3.
5. **`upload_file` validates its argument combinations.** Neither `data` nor `local_file_path` raises `ValueError` instead of `UnboundLocalError`, and `data` without `remote_file_path` raises `ValueError` too — there is no filename to default to, and the old code passed `filename=None` through to the multipart request. Both misbehaviours also ship in 1.0.3.
6. **Every HTTP call carries a timeout**, configurable through a `request_timeout` field (default 120 s) on both tools. Executions are synchronous server-side, so code that legitimately runs longer than the default needs a larger value; the old tools sent no timeout at all and could hang forever on a wedged connection. The field is constrained `gt=0`: Requests rejects a zero or negative timeout with a `ValueError` raised at the first call rather than at the misconfiguration, and there is deliberately no "wait forever" spelling on a public field.
7. **The input sanitizers only strip an interpreter word at a word boundary.** The old regexes matched a bare prefix, so `shopt -s nullglob` became `opt -s nullglob`, `sha256sum` became `a256sum`, and `python_var = 1` became `_var = 1` — with `sanitize_input=True` the default, so every such command was corrupted before it ran.
8. **`auth_headers` raises on a missing token** instead of sending `Authorization: Bearer None` and surfacing later as an opaque 401.

## Deliberately *not* changed

The three API versions in use are preserved, not unified: the Python tool is on `2024-02-02-preview`, the Bash tool on `2025-02-02-preview`, and session delete on `2025-10-02-preview`. Sharing one would silently move a caller to a different service contract, so `api_version` is a required parameter with no default.

`libs/azure-dynamic-sessions` is untouched here — it stays green and releasable until the last PR in this stack deprecates it. The duplication is invisible to users because this package is not published yet.

## Smaller hardening

- `build_session_url` appends the standard parameters last, so a caller-supplied key cannot replace the session identifier.
- The default token provider creates its `DefaultAzureCredential` lazily and reuses it, keeping its HTTP pipeline and token cache instead of rebuilding both on every refresh.
- `_delete_session_async` → `_delete_session_in_background`: it is a daemon thread, not asyncio, and its docstring now records that the delete can be lost at interpreter exit.

## Running the live test

This is the first PR with a live suite, so it also documents how to run one: a `.env.example` template and a README section covering `az login`, `make integration_tests`, and which variable feeds which suite. Both grow in the PRs that add the remaining suites — no PR documents a variable for a test it does not ship.

The template needs `!.env.example` in the root `.gitignore`, because the existing `.env.*` pattern matches it. **The negation is deliberately narrow** — `.env`, `.env.local`, `.env.production` and nested `.env` files all stay ignored, which a bare `!.env*` would not do. Verified with `git check-ignore` against each. The repository already tracks thirteen such templates under `samples/`, every one currently ignored-but-tracked.

## Validation

- 96 unit tests, 100% statement and branch coverage
- the dynamic-sessions end-to-end test against a live session pool

## Notes

The extra requires only `requests` — the tools do not depend on deepagents, which is why `uv.lock` moves by just 7 lines here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@ cover/
 
 .env
 .env.*
+# Templates carry placeholders, not values. Keep this negation narrow: a bare
+# `!.env*` would un-ignore `.env` itself.
+!.env.example
 *.pem
 *.key
 *.crt
--- a/libs/azure-compute/.env.example
+++ b/libs/azure-compute/.env.example
@@ -1,0 +1,8 @@
+# Copy to `.env` and fill in the resources you have. Each integration suite
+# skips itself when its variables are unset, so a partial file runs a partial
+# suite. Authentication is DefaultAzureCredential -- `az login` is enough, and
+# no credential belongs in here.
+
+# tests/integration_tests/dynamic_sessions/test_end_to_end.py
+# A Python-typed session pool.
+AZURE_DYNAMIC_SESSIONS_POOL_MANAGEMENT_ENDPOINT=https://<region>.dynamicsessions.io/subscriptions/<subscription-id>/resourceGroups/<resource-group>/sessionPools/<pool-name>
--- a/libs/azure-compute/README.md
+++ b/libs/azure-compute/README.md
@@ -3,6 +3,37 @@
 LangChain integrations for [Azure Container Apps](https://learn.microsoft.com/azure/container-apps/)
 compute that runs agent-authored code.
 
+## Installation
+
+Install only the integration you need:
+
+```bash
+pip install "langchain-azure-compute[dynamic-sessions]"
+```
+
+Extras control *dependencies*, not which modules are present — every module
+ships in the wheel. Each subpackage checks for its own requirements at import
+time and raises an error naming the extra to install.
+
+## Dynamic sessions
+
+### Tools
+
+`SessionsPythonREPLTool` and `SessionsBashTool` run LLM-authored code in an
+ephemeral session and return its output.
+
+```python
+from langchain_azure_compute.dynamic_sessions import SessionsPythonREPLTool
+
+tool = SessionsPythonREPLTool(pool_management_endpoint="<pool-management-endpoint>")
+print(tool.invoke("print(sum(range(10)))"))
+```
+
+By default a `DefaultAzureCredential` fetches a token for the
+`https://dynamicsessions.io` scope; pass `access_token_provider=` to supply one
+yourself. Requires the `Azure ContainerApps Session Executor` role on the
+session pool.
+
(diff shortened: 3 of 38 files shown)


ShichaoQiu (approved):

ShichaoQiu:
Thanks @darieldatoon , look good to me, have approved the PR

PR #909, opened 2026-08-04
feat(compute): add the Deep Agents dynamic sessions backend

darieldatoon:
# Add the Deep Agents dynamic sessions backend

Part of the stack for #906 · **3 of 6** · base: *Migrate the dynamic sessions tools*

Adds `SessionsBashBackend`, a Deep Agents `SandboxBackendProtocol` implementation over the dynamic-sessions REST API, and pulls deepagents into the `[dynamic-sessions]` extra.

Targets **Shell-typed** session pools: `read`, `write`, `edit`, `ls`, `glob` and `grep` are overridden with shell-native programs instead of the `python3 -c` wrappers `BaseSandbox` uses, which a shell-only pool image may not be able to run.

This is the largest PR in the stack. Two live service behaviours drove much of its shape and are documented below.

## Written against deepagents 0.7 from the start

The equivalent class in `langchain-azure-dynamic-sessions` targets 0.6 and is broken on 0.7: `ls_info`/`glob_info` no longer exist, so both overrides are silently ignored and the base class's `python3 -c` commands run instead; and `read()` returns a `str` where callers expect a `ReadResult`. Neither raises.

Rather than port the old class and then fix it across two PRs, this introduces the backend once, already correct. **There is one implementation to review.**

## Review in this order

| Production | Its tests |
|---|---|
| `backends/_shell.py` — pure command builders and parsers | `test_sessions_shell_programs.py` (runs them under `/bin/sh`) |
| `backends/sessions.py` — the HTTP adapter | `backend/` package (per method), `test_sessions_backend_contract.py` |
| — | `test_backend_live.py` (the same cases against a real Shell pool) |

Three test layers because none alone is sufficient. Contract tests stub `execute()` and check result shapes; the local-shell tests run the generated program in a temp directory; the live tests prove it behaves the same in the pool image. **A mocked transport cannot catch a wrong shell program** — that is how the defects in the old package survived under passing tests.

## Security-relevant invariants

Two rules hold throughout `_shell.py` and should not be relaxed:

- Every interpolated path is bound once via `shlex.quote` and referenced only as `"$p"`. It is never reflected into a double-quoted string, where `$(...)`, backticks and `$VAR` in a filename would be expanded remotely. There is a regression test that writes to a file literally named `$(touch pwned).txt`.
- Caller text — file content, edit strings — travels base64-encoded, never inline.

Errors returned to the model are classified, never interpolated: `requests` embeds the request URL in its messages, which carries the session identifier and the pool endpoint.

## The data plane loses a command's last line of output

Measured against a live Shell pool: the data plane drops a process's final stdout write when that write is immediately followed by exit. **1.6–4% of commands.**

`read` was the only loud symptom, because it alone carries a line count to check its body against. The others were silent:

| Operation | Symptom before this PR |
|---|---|
| `ls` | **empty listing, `error=None`** — a populated directory reported as empty |
| `read` | `read returned no lines` error |
| `glob` | none observed — loop teardown follows its last write |

Every generated program now ends with a **random per-invocation marker**, emitted from an `EXIT` trap so the early error branches carry it too. That fixes as much as it detects: the real output is no longer the final write, so it survives. When the marker is missing anyway, `ls`/`read`/`glob`/`grep` retry once and then report the loss rather than returning a short result as a complete one.

Random rather than a constant: the marker is matched against output that includes file content, so a fixed one could be forged by a file containing it — and would be stripped out of that file's own content.

`write` and `edit` are deliberately **not** gated. Their outcome is the exit status, which the service returns in a separate field and cannot lose with the output; gating them would fail operations that had already succeeded, and retrying them would collide with `write`'s refusal to overwrite. `execute` is also unwrapped, since appending to a caller-supplied command would clobber the exit status it reports — the README tells raw-command users to marker their own output.

A/B against the live pool, 200 rounds, with raw `execute` as the control:

```
execute (unwrapped, control):  5/200  (2.5%)   <- defect still present
ls:                            0/200
glob:                          0/200
read:                          0/200
```

ACA sandboxes was probed identically and is **unaffected** (0/200), so `ACASandbox` in PR 4 needs no equivalent.

## The data plane silently caps output at 4,096 bytes

Also measured live, and deterministic at the exact byte: each execution's stdout and stderr are truncated at 4,096 bytes each. The status stays `"0"` and **no response field marks the cut**.

```
asked 4096 -> stdout 4096   status="0"
asked 4097 -> stdout 4096   status="0"   <- truncated, still "success"
```

It is a total-output cap, not per-write: 125 separate writes totalling 10,000 bytes also come back as 4,096.

`read` therefore pages its window in base64 chunks sized under the cap — the first command returns the line count, the window's byte size and the first chunk; follow-ups page the rest by byte offset. base64 because the transport is a JSON string, and a chunk boundary landing inside a multi-byte character would otherwise be mangled in transit. A window that shrinks between chunks (the file changed) is reported, not stitched.

Chunks are sized against whichever ceiling is lower, the service cap or the caller's `max_output_bytes`, because `execute` applies the latter to every response: a cap under 4 KiB would otherwise slice the read's own base64 record and leave it undecodable. A small cap costs round trips, never a cut record, and one too small to carry even the bookkeeping says which knob to raise. The chunk size is derived from the cap and that bookkeeping rather than tuned by hand, so the two cannot drift apart.

`execute` flags a stream arriving at the cap as `truncated=True`. That is a heuristic — an output of legitimately exactly 4,096 bytes gains a spurious flag — and it is the only signal available, since the service reports none. It is also what routes an over-cap `ls`/`glob`/`grep` into their designed truncation handling: the cap eats the completion marker too, so without the flag the loss machinery answered "output lost; retry it", advice that could never help.

Whether the cap is configurable per pool is an open question for the service team; it does not block this PR.

## Notable implementation choices

- **`edit()` matches over the whole file** with awk `index()`. Counting lines with `grep -cF` is wrong twice: two occurrences on one line read as one and silently take an ambiguous edit, and a multiline `old_string` can never match. `old`/`new` reach awk through `ENVIRON`, not `-v`, which interprets escapes and would corrupt a backslash; their shell variables carry a trailing sentinel because command substitution strips trailing newlines. awk runs under `LC_ALL=C` for byte semantics and so a gawk locale warning on non-UTF-8 content cannot poison the occurrence-count channel.
- **`edit()` matches `old_string` as sent, then as CRLF, then as LF**, applying the same transform to `new_string`, and the file keeps its own line endings. `read` hands the model LF-normalized content, so byte-exact matching broke the ordinary read-then-edit path on any CRLF file. This is what `BaseSandbox.edit` already does.
- **`edit` lands its result with `cat > "$p"`, not `mv`**, preserving inode, mode and hard links, and reports an unknowable occurrence count as `None` rather than guessing.
- **Semantic exit codes are 10+**, disjoint from the small codes shells produce on their own: a failed redirect exits 1, which would otherwise read as "file already exists"; a gawk fatal exits 2, which would read as "appears multiple times".
- **`write`/`edit` payloads above ~90 KB are refused** with an actionable error. Their base64 is inlined into the single `sh -c` program string, which Linux caps at 128 KiB (`MAX_ARG_STRLEN`), and the in-pool failure is an unhelpful E2BIG — measured live at 64 KB passing and 120 KB failing. `edit` budgets `old_string` and `new_string` **together**, since both literals ride in that one argument; 50 KB each passes a per-string check and still generates a 134,931-byte program.
- **Transfers accept only `/mnt/data/<name>`.** The session file API is a flat store, so any other path is rejected with `invalid_path` rather than flattened into the root under a name the caller never asked for. `normpath` runs first, so `/mnt/data/../etc/passwd` cannot traverse out and `/mnt/database` cannot pass as a prefix match.
- **`grep` is overridden.** The base implementation routes slash-containing globs through a `python3 -c` wrapper a Shell pool may not run, and its command carries no completion marker, so the final matches of a search vanished at the measured loss rate. The override is shell-native end to end and supports `max_count`.
- **The async variants delegate to the sync overrides** in a worker thread. `BaseSandbox`'s own async methods rebuild its `python3` commands, so inheriting them would silently bypass every override here — including the `/mnt/data` rules and the loss compensation.
- **`ls` silences the listing's stderr** (a dangling symlink's diagnostic parsed as a directory entry) and honours the error sentinel only on the first line, so a file named `__ERR__…` cannot poison its directory. `read` reports a FIFO or device as `not_a_file` instead of blocking awk until the timeout.
- **`timeout=0` reaches Requests as `None`.** The protocol spells "no timeout" as 0; Requests rejects 0 with a bare `ValueError`, which is not a `RequestException` and escaped the per-file error capture in the transfers.
- **`read()` counts returned lines from the window body** rather than the stripped content. A file containing a single newline is one blank line; counting it as zero produced `end_line=0` against `start_line=1`, which deepagents rejects with a `ValueError`.

## `glob` follows Python's semantics, checked against upstream

`find -path` alone gets the common shapes wrong, so the translation is deliberate: `*` stays within one directory level, `**/name` recurses including the root, a wildcard never matches a leading dot, and the hidden-file exclusion is positional — `.github/**/*.yml` reaches `.github` but not `.github/.private/`, and `**/.gitignore` finds `cfg/.gitignore` without `**` descending into hidden directories.

Also handled: `.` segments are honoured but dropped from the translation (they inflated depth pinning, so `./*.txt` matched nothing); a trailing slash matches directories only; `HEAD/**` reports `HEAD/` itself, Python's zero-expansion form; records are emitted with `printf`, never `echo`, because POSIX `echo` (dash, ash) expands backslash escapes and a filename containing `\t` came back with a literal tab; root failures are probed before `cd` so an unreadable root is distinguishable from a missing one.

Rather than restate upstream's semantics in assertions that drift, the translation is diffed against **`BaseSandbox.glob` itself**. Upstream runs through `execute()`, so a local shell backend serves it as an executable oracle, and the generated primaries are evaluated in-process — no GNU userland needed, so it runs on every platform.

Symlink boundaries follow the base class including where it is asymmetric: `ls` does not pass `-L` (base uses `follow_symlinks=False`) while `glob` keeps following for `is_dir` (base uses `os.path.isdir`), drops a match whose real path leaves the search root, and drops dangling symlinks.

## Known limitations

- `read()` does not classify binary files; everything comes back utf-8 with undecodable bytes replaced. Documented on the method.
- `execute()` is not covered by the completion marker, as above.
- Patterns mixing several `**` segments take a documented `find -path` approximation; `_glob_invocations`' docstring lists every divergence.
- Filenames containing newlines are outside the line-oriented `ls`/`glob`/`grep` contract.
- The generated programs assume GNU userland, which every pool image ships. The unit tests that run them through the local shell are guarded by a probe for `find -printf`, `sort -z` and `xargs -0` and skip where those are absent, so a macOS maintainer sees 44 skips. The skip reason names the missing feature rather than the platform.

## Running the live tests

`AZURE_DYNAMIC_SESSIONS_SHELL_POOL_MANAGEMENT_ENDPOINT` joins `.env.example` and the README's integration-test table. It is a separate variable from the tools' pool on purpose: these tests need a **Shell**-typed pool, and a Python-typed one fails them.

## Validation

- 405 unit tests, 100% statement and branch coverage (44 skip on hosts without GNU userland, e.g. macOS)
- 43 live checks against a Shell-typed session pool

## Notes

`uv.lock` moves because deepagents joins the `[dynamic-sessions]` extra, floored at `0.7.1` — it fixes degenerate read windows (deepagents#5180). `libs/azure-storage` is *not* on the same floor: `main` pins it to `deepagents>=0.6.12,<0.7.0` with a `TODO(#878)`, so the two packages' extras can only co-install once #878 lands.

### DIFF
--- a/libs/azure-compute/.env.example
+++ b/libs/azure-compute/.env.example
@@ -6,3 +6,7 @@
 # tests/integration_tests/dynamic_sessions/test_end_to_end.py
 # A Python-typed session pool.
 AZURE_DYNAMIC_SESSIONS_POOL_MANAGEMENT_ENDPOINT=https://<region>.dynamicsessions.io/subscriptions/<subscription-id>/resourceGroups/<resource-group>/sessionPools/<pool-name>
+
+# tests/integration_tests/dynamic_sessions/test_backend_live.py
+# A Shell-typed session pool. A Python-typed pool will fail these.
+AZURE_DYNAMIC_SESSIONS_SHELL_POOL_MANAGEMENT_ENDPOINT=https://<region>.dynamicsessions.io/subscriptions/<subscription-id>/resourceGroups/<resource-group>/sessionPools/<shell-pool-name>
--- a/libs/azure-compute/README.md
+++ b/libs/azure-compute/README.md
@@ -15,6 +15,9 @@ Extras control *dependencies*, not which modules are present — every module
 ships in the wheel. Each subpackage checks for its own requirements at import
 time and raises an error naming the extra to install.
 
+The Deep Agents backend is marked `@beta`: deepagents is pre-1.0, so this
+interface may change.
+
 ## Dynamic sessions
 
 ### Tools
@@ -33,6 +36,62 @@ By default a `DefaultAzureCredential` fetches a token for the
 `https://dynamicsessions.io` scope; pass `access_token_provider=` to supply one
 yourself. Requires the `Azure ContainerApps Session Executor` role on the
 session pool.
+
+### Deep Agents backend
+
+`SessionsBashBackend` implements the Deep Agents `SandboxBackendProtocol`
+against a **Shell-typed** session pool. Its file operations are bash-native
+rather than the `python3 -c` wrappers `BaseSandbox` uses, which a shell-only
+pool image may not be able to run.
+
+```python
+from deepagents import create_deep_agent
+
+from langchain_azure_compute.dynamic_sessions.backends import (
+    SessionsBashBackend,
+)
+
+backend = SessionsBashBackend(pool_management_endpoint="<pool-management-endpoint>")
+agent = create_deep_agent(model="...", backend=backend)
+```
+
+`upload_files`/`download_files` use the session's file API, which is a flat
+store rooted at `/mnt/data`. Only `/mnt/data/<name>` is storable; any other
+path is rejected with `invalid_path` rather than silently written elsewhere.
+`write()` has no such limit — it goes through the shell and can create any
+path.
+
+The session data plane intermittently drops a command's final line of output
+(measured at 1.6–4%). `ls`, `read`, `glob` and `grep` detect this and retry,
+so they never report a short result as a complete one. **`execute()` is not
+covered**: it runs your command verbatim, and wrapping it would change the
+exit status it returns. If you parse the last line of `execute()` output,
+terminate the command with a marker of your own and re-run when it is missing.
+
+The data plane also caps each command's stdout and stderr at 4,096 bytes,
+silently. `read` pages beneath the cap in base64 chunks, so any file within
+`max_output_bytes` reads intact — larger windows just cost more round trips.
+A listing, glob, or grep whose output exceeds the cap says so: `ls` advises
+listing a subdirectory, `glob`/`grep` return what fits, flagged `truncated`.
+`execute()` sets `truncated=True` on a stream arriving at the cap — a
+heuristic, and the only signal there is, since the service reports none.
+
+Three behavioral notes:
+
+- `write()` refuses to overwrite an existing file, matching the shared
+  `langchain-tests` sandbox suite; the error tells the model to use `edit` or
+  delete first.
+- `write()` content and `edit()` strings ride inside a single shell command,
+  which Linux caps at 128 KiB per string; payloads above ~90 KB are refused
+  with an error saying so. `edit()` budgets `old_string` and `new_string`
+  together, because both travel in that one command. Larger transfers belong
+  in `upload_files`, within its `/mnt/data` limit.
+- `glob()` follows the protocol's Python-glob semantics for the common shapes
+  (`*.py` stays in one directory level, `**/name` recurses); patterns mixing
+  several `**` segments take a documented `find -path` approximation. As in
+  Python, a wildcard does not match a leading dot — pass a pattern that names
+  the hidden entry (`.env`, `.github/**`) to reach one. Naming a hidden
+  directory opens that directory only, not hidden entries beneath it.
 
 ## Development
 
@@ -63,3 +122,4 @@ repository.
 | Variable | Suite | Role |
 |---|---|---|
 | `AZURE_DYNAMIC_SESSIONS_POOL_MANAGEMENT_ENDPOINT` | the tools, against a **Python**-typed pool | `Azure ContainerApps Session Executor` on the pool |
+| `AZURE_DYNAMIC_SESSIONS_SHELL_POOL_MANAGEMENT_ENDPOINT` | `SessionsBashBackend`, against a **Shell**-typed pool | `Azure ContainerApps Session Executor` on the pool |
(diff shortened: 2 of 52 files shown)


ShichaoQiu (approved):

ShichaoQiu:
Thanks @darieldatoon , look good to me, have approved it.

PR #783, opened 2026-06-29
feat(azure-storage): add Deep Agents Azure Blob Storage backend

darieldatoon:
Adds `AzureBlobBackend` (Deep Agents `BackendProtocol` over Azure Blob Storage) to `langchain-azure-storage`, ported from the community `deepagents-azure-blob-backend`. Background in Slack.

```python
from langchain_azure_storage import AzureBlobBackend, AzureBlobConfig
```

## Worth a close look

- **Packaging** — new `[deepagents]` extra gated to `python_version >= "3.11"` (deepagents requires 3.11+). Note that the main package is scoped to Python 3.10+, but the deepagents package requires 3.11+. If we choose to drop support for 3.10, then we can simplify some of this logic.
- **CI** — `_test.yml` gains an Azurite step scoped to the storage job on 3.12 (with a readiness probe); the other libs are untouched. Integration + contract suites run there.
- **Shared user-agent** — factored `azpartner-langchain/<version>` into `_user_agent.py`; the loader and backend now both tag requests with it.
- **maintainers.md** — @darieldatoon as DRI for the backend namespace.

## Testing

ruff + mypy clean on 3.10/3.12; unit tests pass (deepagents tests skip on 3.10); integration + contract pass against Azurite.

PR #910, opened 2026-08-04
feat(compute): add ACASandbox deep agents backend

darieldatoon:
# Add `ACASandbox`

Part of the stack for #906 · **4 of 6** · base: *Add the Deep Agents dynamic sessions backend*

Adds `ACASandbox`, a Deep Agents `SandboxBackendProtocol` implementation over an Azure Container Apps sandbox (`Microsoft.App/sandboxGroups`), behind a new `[sandboxes]` extra.

**A different Azure product from dynamic sessions**, not a tier of it — different ARM resource type, different data plane. The README table covers choosing between them.

It wraps a client you construct, so it never builds endpoints or handles credentials itself, and you keep control of the sandbox lifecycle through that same client.

## Review in this order

| Production | Its tests |
|---|---|
| `sandboxes/_results.py` — pure SDK→protocol conversions | `sandbox/test_results.py`, `sandbox/test_read.py` |
| `sandboxes/sandbox.py` — the adapter | `sandbox/test_execute.py`, `test_write.py`, `test_upload_download.py`, `test_errors.py`, `test_async.py` |
| `sandboxes/__init__.py` — extras import guards | `test_import_guards.py`, `test_beta_markers.py` |

## Notable choices

- **`read`/`write` go through the SDK, not the shell.** `BaseSandbox.read()` pipes content through `execute()`, which can hang or exceed transport limits on large files. `BaseSandbox.write()` runs a `python3 -c` mkdir preflight and then delegates content to `upload_files` — already this class's SDK path — so the override skips the preflight and adds the overwrite refusal.
- **The async overrides are narrower than they look.** Only `BaseSandbox.aread` truly rebuilds the shell path through `aexecute`; `awrite`/`aupload_files`/`adownload_files` in the base delegate to the sync methods on a thread, which are already this class's SDK calls. The overrides exist to use a supplied `async_client`; without one they offload the sync SDK call to a thread themselves, so both configurations behave identically. The sync/async pairs are forked rather than bridged, per the direction on #783.
- **`read`/`aread` stat first.** Oversized binaries are refused without a download; files over 10 MiB fall back to the base class's server-side pagination, because the SDK read returns the whole body and one page would otherwise cost the whole file on every page; a directory is reported as one.
- **Batch transfers catch `AzureError`, not `HttpResponseError`.** `ServiceRequestError` is not an `HttpResponseError`, so a transport blip on one file would otherwise abort the whole batch despite the documented partial-success contract.
- **`execute`/`aexecute` catch `AzureError`** and return an error `ExecuteResponse` instead of raising into the agent loop. The common cause is a command outrunning the transport's ~300 s read timeout.
- **A 404 is classified by operation, not by exception type.** Reading or downloading a path that is not there is `file_not_found`; a 404 from `exec`, a write or an upload is not — a command has no file to miss, and an upload's destination is *expected* not to exist. Reporting those as `file_not_found` sends the model into file recovery for what is actually an unavailable sandbox. The dynamic-sessions transfer classifier draws the same line.
- **Errors are classified, never interpolated.** azure-core embeds response bodies in its messages and these strings reach the agent. Failures that cannot be mapped report status or exception type rather than borrowing a protocol code, since a wrong code sends the agent down the wrong recovery path. Typed checks and the status code outrank substring sniffing, so a message that merely *contains* "denied" cannot steer the agent into permission recovery.
- **The write refusal distinguishes a directory** and tells the model to use `edit` or delete first, since the deepagents middleware's tool description promises overwrite.
- **`@beta`**, matching `AzureBlobBackend` and the Azure document loaders: deepagents is pre-1.0 and ACA sandboxes is Early Access.

`_results.py` vendors the extension-based text/binary classifier with a parity test against the installed deepagents — the upstream helper is private and could move.

The import guard survives `find_spec` *raising* for the missing parent namespace; the curated install-the-extra message was otherwise unreachable in exactly the situation it exists for.

## Known limitation

`SandboxClient.exec()` has no server-side command timeout, so `execute(..., timeout=N)` wraps the command in coreutils `timeout`. On an image without it the command still runs, untimed. Documented rather than enforced; returning an unsupported-capability error would break minimal and BYO images that work today. The exit-124 "timed out" note only appears when the wrapper was actually applied.

Command timeouts above the transport's ~300 s read timeout are unreachable on the SDK default; the class docstring and README say so and point at constructing the client with a longer-read-timeout transport.

## Validation

- 521 unit tests, 100% statement and branch coverage (44 skip on hosts without GNU userland, e.g. macOS)

Live contract evidence is the next PR, which runs the shared standard suite against a real sandbox.

## Notes

`uv.lock` grows because the `[sandboxes]` extra adds `azure-containerapps-sandbox` — a preview SDK that documents itself as subject to breaking change, hence the tight `<0.2` ceiling.


ShichaoQiu (approved):

ShichaoQiu:
Thanks @darieldatoon , look good to me, have approved it

PR #911, opened 2026-08-04
test(compute): run the standard sandbox suite against ACASandbox

darieldatoon:
# Run the shared sandbox standard suite against `ACASandbox`

Part of the stack for #906 · **5 of 6** · base: *Add `ACASandbox`*

Subclasses `SandboxIntegrationTests` from `langchain-tests` and runs it against a real Azure Container Apps sandbox.

Unlike the blob backend in `langchain-azure-storage`, which had to fork the suite because it implements the filesystem-only `BackendProtocol`, `ACASandbox` is a genuine sandbox and can subclass it directly.

This is the acceptance evidence for PR 4: it exercises protocol edge cases upstream has accumulated — unicode, escaped content, large payloads, pagination — that no local suite here covers.

## Why it is separate

Test-only, and it adds a dependency (`langchain-tests`) the package does not otherwise need. **Fold it into PR 4 if these branches may merge independently**, since the backend should not land without its contract evidence. Kept separate on the assumption the stack lands as a unit.

## Cost and cleanup

One sandbox is created per class and deleted afterwards, so these consume real Azure resources. They carry an `aca_sandbox` marker: run only them with `-m aca_sandbox`, or skip them with `-m "not aca_sandbox"`. The README documents the marker so a reviewer can skip the billable suite:

```bash
uv run --frozen --all-extras --group test --group test_integration \
    pytest tests/integration_tests -m "not aca_sandbox"
```

The fixture is built on `ExitStack` so the credential and both clients close even when sandbox creation itself raises.

Cleanup is keyed on a unique run label and registered **before** the create call. `begin_create_sandbox` issues its PUT synchronously, so a failure while waiting for Running — poller timeout, terminal `Failed`, Ctrl-C — happens with the sandbox already existing server-side. A delete callback registered only after `.result()` would never run on exactly those paths, stranding a billable sandbox. Deletion goes through `list_sandboxes(labels=...)`, which covers the success path and every failure path alike; when the PUT never landed the listing is simply empty.

`python-3.12` rather than `ubuntu` as the image: the suite runs `python -c ...`, and the `ubuntu` preset provides only `python3`.

## Running the live tests

The four `AZURE_CONTAINER_APPS_*` variables join `.env.example` and the README's integration-test table here rather than in PR 4, because this is the layer that first reads them — `ACASandbox` itself ships with no live suite.

## Validation

- 131 integration tests against live Azure: 86 from this suite, 44 dynamic-sessions (43 backend contract checks plus the end-to-end tool test), and 1 compile placeholder
- 521 unit tests, 100% coverage, ruff + mypy clean (unchanged from PR 4)
- `list_sandboxes` reports 0 remaining after every run

## Notes

`uv.lock` grows because `langchain-tests` joins the `test_integration` dependency group, marker-gated to Python >= 3.11.

A residual `PytestRemovedIn10Warning` in the live run comes from `SandboxIntegrationTests`' own class-scoped fixture in `langchain-tests` (an instance-method fixture), not from this adapter — ours uses the recommended `@classmethod` form. Worth an upstream note; not fixable here.

### DIFF
--- a/libs/azure-compute/.env.example
+++ b/libs/azure-compute/.env.example
@@ -10,3 +10,10 @@ AZURE_DYNAMIC_SESSIONS_POOL_MANAGEMENT_ENDPOINT=https://<region>.dynamicsessions
 # tests/integration_tests/dynamic_sessions/test_backend_live.py
 # A Shell-typed session pool. A Python-typed pool will fail these.
 AZURE_DYNAMIC_SESSIONS_SHELL_POOL_MANAGEMENT_ENDPOINT=https://<region>.dynamicsessions.io/subscriptions/<subscription-id>/resourceGroups/<resource-group>/sessionPools/<shell-pool-name>
+
+# tests/integration_tests/sandboxes/test_standard.py
+# All four are required together. These provision and delete a real sandbox.
+AZURE_SUBSCRIPTION_ID=00000000-0000-0000-0000-000000000000
+AZURE_CONTAINER_APPS_RESOURCE_GROUP=<resource-group>
+AZURE_CONTAINER_APPS_SANDBOX_GROUP=<sandbox-group>
+AZURE_CONTAINER_APPS_REGION=<region>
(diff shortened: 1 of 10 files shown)


ShichaoQiu (approved):

ShichaoQiu:
Thanks @darieldatoon , look good to me, have approved it

PR #912, opened 2026-08-04
chore(dynamic-sessions): deprecate in favour of langchain-azure-compute

darieldatoon:
# Deprecate `langchain-azure-dynamic-sessions`

Part of the stack for #906 · **6 of 6** · base: *Run the shared sandbox standard suite against `ACASandbox`* · top of stack

Points users at `langchain-azure-compute`. Placed **last** in the stack so the whole replacement package exists first: the warning and the README can then describe what users are actually moving to — both integrations, not a package still being assembled.

**Touches `libs/azure-dynamic-sessions` only.** No changes to the new package.

## What it does

- Importing the package emits a `DeprecationWarning` naming the replacement and the import rewrite.
- The README leads with the migration, using an absolute GitHub URL — a repo-relative link does not resolve in the rendered PyPI description. The diff maps all three public names.
- The `description` now leads with "Deprecated: use langchain-azure-compute". The **version bump and its `## Changelog` entry are deliberately deferred** to a separate release PR, per review — this PR carries the deprecation only, so `pyproject.toml` stays at 1.0.3 and `uv.lock` is untouched.
- A separate `RuntimeWarning` when the installed deepagents is >= 0.7, where this package's backend is silently broken: its file-operation overrides are ignored and `read()` returns the wrong type. Neither raises, so an affected agent degrades quietly.

Warning rather than raising, so an upgrade whose only purpose is to deprecate does not break working setups.

## Worth a close look

The version check matches leading `major.minor` digits rather than splitting on dots and calling `int()`. The old form raised on any prerelease and so skipped the warning for exactly the versions most likely to hit this.

Comparing release numbers rather than PEP 440 order is deliberate: `Version("0.7rc1")` sorts *below* `Version("0.7")`, so an ordering comparison would also let 0.7 prereleases through. `tests/unit_tests/test_deprecation.py` enforces this across `0.6.12`, `0.7.0`, `0.7rc1`, `0.7.0.dev0`, `0.8.1`, `1.0.0`, a non-version string and a missing distribution — deepagents stand-ins are injected into `sys.modules`, since this package deliberately does not depend on it — and asserts the `DeprecationWarning` on a fresh import.

One visibility note: `DeprecationWarning` is hidden by Python's default filters outside `__main__` and test runners. That is the convention for library deprecations and is chosen deliberately over a louder `UserWarning`.

## Deliberately not fixed

The underlying 0.7 bugs. The fix ships in the new package and this one is going away.

## Validation

- 49 unit tests for the old package (11 new for the warnings), ruff and mypy clean — run on **both** CI matrix legs, Python 3.10 and 3.14
- `uv lock --check` passes under CI's pinned uv 0.12.3; `uv.lock` is byte-identical to `main`

## Notes

Rebased onto current `main`, which moved the CI matrix from Python 3.12 to 3.14 and uv from 0.11.2 to 0.12.3; both legs verified locally. `uv.lock` is no longer modified by this PR at all.

### DIFF
--- a/libs/azure-dynamic-sessions/README.md
+++ b/libs/azure-dynamic-sessions/README.md
@@ -1,5 +1,33 @@
 # langchain-azure-dynamic-sessions
 
+> [!WARNING]
+> **This package is deprecated and will not receive further fixes.** It has
+> been superseded by
+> [`langchain-azure-compute`](https://github.com/langchain-ai/langchain-azure/tree/main/libs/azure-compute),
+> which carries the dynamic sessions integrations forward under the wider
+> Azure Container Apps umbrella — and additionally covers Azure Container Apps
+> sandboxes, the stateful counterpart to dynamic sessions.
+>
+> ```bash
+> pip install "langchain-azure-compute[dynamic-sessions]"
+> ```
+>
+> Then update imports:
+>
+> ```diff
+> - from langchain_azure_dynamic_sessions import SessionsPythonREPLTool
+> + from langchain_azure_compute.dynamic_sessions import SessionsPythonREPLTool
+> - from langchain_azure_dynamic_sessions import SessionsBashTool
+> + from langchain_azure_compute.dynamic_sessions import SessionsBashTool
+> - from langchain_azure_dynamic_sessions.backends import SessionsBashBackend
+> + from langchain_azure_compute.dynamic_sessions.backends import SessionsBashBackend
+> ```
+>
+> Migrating also fixes `SessionsBashBackend` on **deepagents >= 0.7**, where it
+> is broken here: its file-operation overrides are silently ignored and
+> `read()` returns the wrong type. Neither failure raises, so an affected agent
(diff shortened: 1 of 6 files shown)


ShichaoQiu:
We can update the version in a separate PR and we also need to update the change log in readme file in that PR.

darieldatoon:
Thanks @ShichaoQiu — done. The version bump is out of this PR (`pyproject.toml` stays at 1.0.3 and `uv.lock` is no longer touched), so this now carries the deprecation only. The bump and its `## Changelog` entry will follow in a separate release PR.

ShichaoQiu (approved):

ShichaoQiu:
Thanks @darieldatoon , look good to me, I have approved the PR. Feel free to merge it if there is no other update or issue. Thanks a lot.

darieldatoon:
Thanks, @ShichaoQiu! Let me know when a new release has been cut and the new package name is available!

PR #952, opened 2026-08-18
refact: Add flexible version management inputs for UV and Python

Wixee:
Introduce `uv-version` and `python-versions` inputs to workflows, allowing for more flexible management of UV and Python versions during CI processes. This change enhances the configurability of the workflows.


santiagxf (approved):
Just take the `uv.lock` file out of the PR.

Issue #815, opened 2026-07-07
[tracking] Revisit Python 3.10 support / deepagents `>=3.11` marker at 3.10 EOL


darieldatoon:
## Decision (resolved)

We are **keeping the `langchain-azure-storage` (and repo) minimum at Python 3.10** and living with the import-time workaround for the `[deepagents]` extra. We'll **revisit bumping to 3.11 near Python 3.10 EOL (October 2026)** or whenever `langchain` core drops 3.10 support — whichever comes first. This issue tracks that revisit.

Decided with the LangChain Azure package owners as a follow-up to #783; the driver is that moving the whole Azure package set off 3.10 now costs more (dependency sprawl + shared-CI/workflow rework across all Azure packages) than the benefit, while 3.10 EOL is only ~3–4 months out. New work is generally pinned to 3.11, but this bump is deferred until necessary.

## Background

`deepagents` requires Python `>=3.11`, but `langchain-azure-storage` targets `>=3.10` to stay in lock step with `langchain` core (which still supports 3.10). The `[deepagents]` extra therefore pulls a 3.11-only dependency into a 3.10-supporting package.

Current implementation keeps a `python_version >= "3.11"` environment marker on the `deepagents` dependency. On 3.10 the extra resolves to nothing, so importing the backend hits the subpackage's `ImportError` shim — the same shim that guards a missing extra on any Python version. This works and CI is green. The downside is that on 3.10 an unsupported install fails at import time rather than **install time**, which is the usual expectation for an extra — the marker leaks a tooling limitation into the public interface.

## Why we can't just drop the marker today

`uv` resolves a single universal lockfile across the whole `requires-python` range:

- **Drop the marker** (keep `requires-python = ~=3.10`) → `uv lock` fails: the 3.11-only `deepagents` makes the resolution unsatisfiable for 3.10 (also breaks `uv lock --check` / `uv sync`).
- **Restrict the lock to `>=3.11`** (`[tool.uv] environments = ["python_version >= '3.11'"]`) → the lock resolves, but drops 3.10 entirely, so the 3.10 document-loader CI jobs can't sync.

So we can't get an install-time error on 3.10 while keeping both a working universal lock and 3.10 support for the loaders. Relevant `uv` work: astral-sh/uv#13735 (dependency-groups support for this, but not extras yet).

## When we revisit (checklist)

At 3.10 EOL / when `langchain` core drops 3.10, bump the package minimum to 3.11 and:

[...]

PR #995, opened 2026-09-01
Deprecate Python 3.10 support for `langchain-sqlserver`

miljann995:
## Summary

Raise the minimum supported Python version for `langchain-sqlserver` from
Python 3.10 to Python 3.11.

This change is scoped to the SQL Server package. Other packages in this
monorepo continue to support and validate against Python 3.10.

## Motivation

Python 3.10 is approaching end of life in October 2026. Since
`langchain-sqlserver` is independently packaged and does not have runtime
dependencies on sibling packages in this repository, its minimum version can
be raised without changing the support policy of those packages.

Declaring Python 3.11 as the package minimum also ensures that package managers
reject unsupported Python 3.10 environments at install time.

Related to #815, which tracks the broader repository-wide Python 3.10
retirement. This PR deliberately limits the change to `langchain-sqlserver`.

## Changes

- Set `langchain-sqlserver` `requires-python` to `>=3.11,<4.0`.
- Update its universal lockfile to cover Python `>=3.11,<4.0`.
- Document the Python 3.11 requirement in the package README.
- Run SQL Server lint, type-check, unit-test, and integration-compilation jobs
  on Python 3.11 and 3.14.
- Preserve the default Python 3.10 and 3.14 CI matrix for sibling packages.
- Build the SQL Server Test PyPI candidate on Python 3.11 while retaining
  Python 3.10 for sibling packages.
- Update repository agent guidance to describe the package-specific version
  policy.

Production artifacts were already built with Python 3.11, so the production
build interpreter does not need to change.

## Compatibility impact

New `langchain-sqlserver` releases produced from this change cannot be installed
on Python 3.10. Package managers may still select an older compatible release
when no package version is explicitly requested.

There are no runtime API or behavior changes.

## Validation

Validated on Python 3.11 and Python 3.14 using the CI-pinned uv 0.12.3:

- `uv lock --check`
- Ruff checks and formatting verification
- mypy for package and test code
- Unit tests: 17 passed, 1 skipped on each Python version
- Integration-test compilation: 1 passed, 99 live tests deselected on each version
- Wheel and source distribution build
- Wheel import on Python 3.11
- Wheel metadata accepts Python 3.11/3.14 and rejects Python 3.10

Live SQL Server integration tests were not run because the required connection
environment variables were not available.

### DIFF
--- a/.github/agents/langchain-azure.agent.md
+++ b/.github/agents/langchain-azure.agent.md
@@ -61,7 +61,7 @@ make format && make lint_package && make lint_tests
 
 ### CI/CD
 
-CI is path-aware — it only runs lint/test for packages with changed files (via `.github/scripts/check_diff.py`). Tests run on Python 3.10 and 3.12. Infrastructure changes (`.github/workflows`, `.github/scripts`) trigger all packages.
+CI is path-aware — it only runs lint/test for packages with changed files (via `.github/scripts/check_diff.py`). Tests run on Python 3.10 and 3.14 for all packages except for `langchain-sqlserver` that runs on Python 3.11 and 3.14. Infrastructure changes (`.github/workflows`, `.github/scripts`) trigger all packages.
 
 The main CI workflow (`.github/workflows/check_diffs.yml`) fans out into:
 - `_lint.yml` — `uv lock --check`, `make lint_package`, `make lint_tests`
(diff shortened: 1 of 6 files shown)


santiagxf:
Addressed in #1021

PR #1017, opened 2026-09-07
fix(azure-ai): URL-encode query and filter in AzureAISearchRetriever

### LINKED ISSUE #1016: AzureAISearchRetriever does not URL-encode the search query or filter
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes the issue.
- [x] I searched for similar issues and didn't find a duplicate.
- [x] I am sure this is a bug in a langchain-azure package rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of the package.

### Package

langchain-azure-ai

### Component

Retrievers (AzureAISearchRetriever)

### Reproduction Steps / Example Code

`AzureAISearchRetriever._build_search_url` builds the REST request URL by
directly interpolating the raw query string (and the `filter` option) into
the URL, with no percent-encoding:

```python
endpoint_path = f"indexes/{self.index_name}/docs?api-version={self.api_version}"
top_param = f"&$top={self.top_k}" if self.top_k else ""
filter_param = f"&$filter={self.filter}" if self.filter else ""
return base_url + endpoint_path + f"&search={query}" + top_param + filter_param
```

Any query that contains a character with meaning in a URL query string (`&`,
`#`, `%`, `+`, ...) breaks:

```python
from langchain_azure_ai.retrievers.azure_ai_search import AzureAISearchRetriever

retriever = AzureAISearchRetriever(
    service_name="my-svc", index_name="my-index", api_key="k"
)
print(retriever._build_search_url("cats & dogs"))
# https://my-svc.search.windows.net/indexes/my-index/docs?api-version=2023-11-01&search=cats & dogs
```

`requests`/`aiohttp` treat that `&` as a parameter delimiter, so the actual
search sent to Azure AI Search is just `cats` — the rest of the query is
silently dropped (and turned into a stray, malformed query-string entry).

It also means the search text can inject extra query parameters, including a
second `$filter`:

```python
print(retriever._build_search_url("foo&$filter=1 eq 1"))
# ...&search=foo&$filter=1 eq 1
```

If an application uses `filter=` for access control (e.g. restricting
results to a tenant or user), unescaped user search input can append its own
`$filter` onto the request.

### Error Message and Stack Trace (if applicable)

_No response — this doesn't raise, it silently sends a different query than
the one passed to the retriever._

### Description

* I'm using `AzureAISearchRetriever` with a search query that contains an
  ampersand, percent sign, or similar character (e.g. "R&D", "100% organic").
* I expect the full query text to be sent to Azure AI Search.
* Instead, the query gets truncated at the special character (or otherwise
  corrupted), because it's concatenated into the request URL without being
  percent-encoded. The same applies to the `filter` constructor argument.

### System Info

langchain-azure-ai (main branch)

cr-sbarbouche:
### Description

`AzureAISearchRetriever._build_search_url` builds the REST request URL by
concatenating the raw search query (and the `filter` option) straight into
the query string:

```python
filter_param = f"&$filter={self.filter}" if self.filter else ""
return base_url + endpoint_path + f"&search={query}" + top_param + filter_param
```

Neither value is percent-encoded. `requests`/`aiohttp` treat a literal `&`
in the assembled URL as a parameter delimiter, so any query containing one
(`"cats & dogs"`, `"R&D"`) gets silently truncated at the `&` — only `cats`
is ever sent to Azure AI Search. The same interpolation also lets query
text such as `"foo&$filter=1 eq 1"` append an unintended `$filter` parameter
to the request.

Closes #1016.

### Fix

Percent-encode `query` and `filter` with `urllib.parse.quote` before they're
interpolated into the URL, the same way any REST client builds a URL from
user-controlled text. No new dependency — `urllib.parse` is stdlib.

### Testing

Added `tests/unit_tests/test_azure_ai_search_retriever.py` (this retriever
previously had no unit test coverage, only a quarantined integration test
that requires a live service):

- Direct tests on `_build_search_url` covering `&`, `%`/`#`, and the
  parameter-injection case, plus filter encoding.
- `_get_relevant_documents` / `_aget_relevant_documents` tests (mocking
  `requests.get` and `aiohttp.ClientSession`) asserting the full query
  reaches the outgoing request and documents are parsed correctly.

All 6 new tests fail against the pre-fix code (confirmed locally) and pass
with the fix. Full unit suite, `ruff check`, `ruff format --diff`, `mypy`,
and `codespell` are clean.

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/retrievers/azure_ai_search.py
+++ b/libs/azure-ai/langchain_azure_ai/retrievers/azure_ai_search.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import json
 from typing import Any, Dict, List, Optional
+from urllib.parse import quote
 
 import aiohttp
 import requests
@@ -173,8 +174,9 @@ class AzureAISearchRetriever(BaseRetriever):
             base_url = self.service_name
         endpoint_path = f"indexes/{self.index_name}/docs?api-version={self.api_version}"
         top_param = f"&$top={self.top_k}" if self.top_k else ""
-        filter_param = f"&$filter={self.filter}" if self.filter else ""
-        return base_url + endpoint_path + f"&search={query}" + top_param + filter_param
+        filter_param = f"&$filter={quote(self.filter, safe='')}" if self.filter else ""
+        search_param = f"&search={quote(query, safe='')}"
+        return base_url + endpoint_path + search_param + top_param + filter_param
 
     @property
     def _headers(self) -> Dict[str, str]:
--- a/libs/azure-ai/tests/unit_tests/test_azure_ai_search_retriever.py
+++ b/libs/azure-ai/tests/unit_tests/test_azure_ai_search_retriever.py
@@ -1,0 +1,98 @@
+"""Unit tests for AzureAISearchRetriever."""
+
+import json
+from unittest.mock import AsyncMock, MagicMock, patch
+from urllib.parse import parse_qs, urlparse
+
+from langchain_azure_ai.retrievers.azure_ai_search import AzureAISearchRetriever
+
+
+def make_retriever(filter: str | None = None) -> AzureAISearchRetriever:
+    return AzureAISearchRetriever(
+        service_name="my-svc", index_name="my-index", api_key="k", filter=filter
+    )
+
+
+def test_build_search_url_preserves_query_with_ampersand() -> None:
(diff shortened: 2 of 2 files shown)


cr-sbarbouche:
@santiagxf mind taking a look? This is an isolated fix in the AzureAISearchRetriever REST path, unrelated to the AzureSearch vector store.

santiagxf:
LGTM!

PR #1025, opened 2026-09-09
chore: Refine dependency group

Wixee:
Refine langchain-azure-ai dependency groups


libs/azure-ai/pyproject.toml
```diff
@@ -39,6 +39,8 @@ dependencies = [
   "azure-search-documents>=11.4,<13.0",
   "langchain>=1.2.12,<2.0.0",
   "langchain-openai>=1.0.0,<2.0.0",
+  "anthropic>=0.111.0,<1.0",
```
a1exwang, line 42: why making it a hard dependency?

Wixee: Just to ensure that `AzureAIAnthropicChatModel` loads its dependencies in the same way as `AzureAIOpenAIApiChatModel`. So that user can use this class without any other effort.

a1exwang (approved):

PR #1030, opened 2026-09-12
fix(azure-ai): scope semantic cache lookup/clear by llm_string

### LINKED ISSUE #1029: AzureCosmosDBMongoVCoreSemanticCache lookup/clear are not scoped by llm_string
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes the issue.
- [x] I searched for similar issues and didn't find a duplicate.
- [x] I am sure this is a bug in a langchain-azure package rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of the package.

### Package

langchain-azure-ai

### Component

Other

### Authentication Method

Not applicable

### Related Issues / PRs

_None found._

### Reproduction Steps / Example Code

`AzureCosmosDBMongoVCoreSemanticCache` is a `BaseCache` implementation meant to be
installed once via `set_llm_cache(...)` and shared by every LLM/chat-model instance
in a process. LangChain calls `lookup(prompt, llm_string)` / `update(prompt,
llm_string, return_val)` on it, where `llm_string` is a serialization of the calling
model's class + parameters — the whole point being that a cached generation from one
model/config is never handed back for a different one.

The cache buckets entries by `llm_string` through `_get_llm_cache`, which builds a
distinct `index_name` (`f"cache:{_hash(llm_string)}"`) per `llm_string`:

```python
# langchain_azure_ai/vectorstores/cache.py
def _get_llm_cache(self, llm_string: str) -> AzureDocumentDBVectorSearch:
    index_name = self._index_name(llm_string)
    ...
    if self.cosmosdb_client:
        collection = self.cosmosdb_client[self.database_name][self.collection_name]
        self._cache_dict[index_name] = AzureDocumentDBVectorSearch(
            collection=collection,
            embedding=self.embedding,
            index_name=index_name,
        )
```

but every one of these `AzureDocumentDBVectorSearch` instances is handed the *same*
underlying `collection` (`database_name`/`collection_name` are fixed for the whole
cache). `index_name` only labels the `cosmosSearch` index definition used at
create-time — the aggregation pipeline built in
`_get_pipeline_vector_ivf`/`_hnsw`/`_diskann` (in `langchain_azure_cosmosdb`) doesn't
even reference an index name at query time, so it searches every document in the
collection, regardless of which `llm_string` wrote it.

`lookup()` then calls `similarity_search` with no filter at all:

```python
def lookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
    llm_cache = self._get_llm_cache(llm_string)
    results = llm_cache.similarity_search(
        query=prompt,
        k=1,
        kind=self.kind,
        ef_search=self.ef_search,
        l_search=self.l_search,
        score_threshold=self.score_threshold or 0.0,
        oversampling=self.oversampling,
    )
    ...
```

So if two model configurations share the same `database_name`/`collection_name`
(a completely normal setup — one Cosmos DB collection backing the cache for
several deployments/models in an app), a lookup for one model can return a
generation that was actually produced — and cached — by a different model.

`clear()` has the same problem the other way around: it deletes every document in
the collection, not just the ones for the given `llm_string`:

```python
def clear(self, **kwargs: Any) -> None:
    index_name = self._index_name(kwargs["llm_string"])
    if index_name in self._cache_dict:
        self._cache_dict[index_name].get_collection().delete_many({})
```

Concretely (pseudo-code, mirrors the unit test I'll attach to the fix):

```python
cache = AzureCosmosDBMongoVCoreSemanticCache(
    cosmosdb_connection_string=CONN,
    database_name="db", collection_name="coll",
    embedding=embeddings,
)
cache.update("what is 2+2?", "gpt-4o|temperature=0", [Generation(text="4")])

# A different deployment, sharing the same collection:
cache.lookup("what is 2+2?", "gpt-4o-mini|temperature=1")
# -> returns the gpt-4o entry instead of a cache miss

cache.clear(llm_string="gpt-4o-mini|temperature=1")
# -> wipes the gpt-4o entry too, even though it was never asked to
```

### Error Message and Stack Trace (if applicable)

_No response — this doesn't raise, it silently returns/deletes the wrong cache
entries._

### Description

* I'm using `AzureCosmosDBMongoVCoreSemanticCache` as a shared semantic cache for
  more than one Azure OpenAI/Azure AI model deployment pointed at the same Cosmos
  DB database/collection.
* I expect `lookup(prompt, llm_string)` to only ever return a generation that was
  `update()`-d under that exact `llm_string`, and `clear(llm_string=...)` to only
  remove that model's entries.
* Instead, `lookup` can return another model's cached answer (silently, as a
  normal cache hit), and `clear` for one model empties the cache for every model
  sharing the collection.

### System Info

langchain-azure-ai (main branch), `langchain_azure_ai/vectorstores/cache.py`

cr-sbarbouche:
### Description

`AzureCosmosDBMongoVCoreSemanticCache` is meant to bucket cache entries by
`llm_string` — a serialized model class + params — so a generation cached for
one model/config is never handed back as a hit for a different one. It builds
a distinct `index_name` per `llm_string` in `_get_llm_cache`, but every one of
those `AzureDocumentDBVectorSearch` instances wraps the *same* underlying
Cosmos DB collection (`database_name`/`collection_name` are fixed for the
whole cache). The `cosmosSearch` aggregation pipeline doesn't take an index
name at query time, so `index_name` never actually partitions documents —
it's only used for index bookkeeping (`index_exists`/`create_index`).

`lookup()` called `similarity_search` with no filter at all, so it searched
every document ever cached in the collection and could return a generation
that a *different* model wrote. `clear(llm_string=...)` had the mirror bug:
it called `delete_many({})`, wiping the entire collection instead of just the
entries for that `llm_string`.

This only bites when a single Cosmos DB database/collection backs the cache
for more than one model or configuration, which is a completely ordinary
setup — one cache shared by a couple of Azure OpenAI/Azure AI deployments,
or by the same deployment at more than one temperature.

Closes #1029.

### Fix

Filter both operations on `metadata.llm_string`: `lookup()` now passes
`pre_filter={"metadata.llm_string": llm_string}` to `similarity_search`
(already a supported parameter on `AzureDocumentDBVectorSearch`, just never
wired up here), and `clear()` passes the same filter to `delete_many`
instead of an empty one. No new dependency, no interface change.

### Testing

Added `tests/unit_tests/test_semantic_cache.py` — this class had no unit
test coverage before, only an integration test requiring a live Cosmos DB
instance. `langchain_azure_ai.vectorstores.cache` imports
`langchain_azure_cosmosdb` at module scope, so the tests register a small
stub for it in `sys.modules` before importing (same pattern already used in
this suite for other optional SDKs, e.g. `test_speech_to_text_tool.py`). The
stub's fake vector store mirrors the real bug's shape: every `llm_string`
shares one in-memory collection, and results are only excluded when an
explicit filter says so — exactly like the real `cosmosSearch`/`delete_many`
filters.

- `test_lookup_does_not_return_a_different_llm_strings_entry` — a lookup
  under one `llm_string` must not see an entry written under another.
- `test_clear_only_removes_entries_for_the_given_llm_string` — clearing one
  model's cache must leave another model's entry intact.
- `test_update_then_lookup_round_trips_for_the_same_llm_string` — sanity
  check that the ordinary single-model path still works.

Confirmed the first two fail against the pre-fix code and pass with the fix.
Full unit suite, `ruff check`, `ruff format --diff`, `mypy`, and `codespell`
are clean.

### DIFF
--- a/libs/azure-ai/langchain_azure_ai/vectorstores/cache.py
+++ b/libs/azure-ai/langchain_azure_ai/vectorstores/cache.py
@@ -323,6 +323,7 @@ class AzureCosmosDBMongoVCoreSemanticCache(BaseCache):
             l_search=self.l_search,
             score_threshold=self.score_threshold or 0.0,
             oversampling=self.oversampling,
+            pre_filter={"metadata.llm_string": llm_string},
         )
         if results:
             for document in results:
@@ -361,9 +362,12 @@ class AzureCosmosDBMongoVCoreSemanticCache(BaseCache):
 
     def clear(self, **kwargs: Any) -> None:
         """Clear semantic cache for a given llm_string."""
-        index_name = self._index_name(kwargs["llm_string"])
+        llm_string = kwargs["llm_string"]
+        index_name = self._index_name(llm_string)
         if index_name in self._cache_dict:
-            self._cache_dict[index_name].get_collection().delete_many({})
+            self._cache_dict[index_name].get_collection().delete_many(
+                {"metadata.llm_string": llm_string}
+            )
 
     @staticmethod
     def _validate_enum_value(value: Any, enum_type: Type[Enum]) -> None:
--- a/libs/azure-ai/tests/unit_tests/test_semantic_cache.py
+++ b/libs/azure-ai/tests/unit_tests/test_semantic_cache.py
@@ -1,0 +1,198 @@
+"""Unit tests for AzureCosmosDBMongoVCoreSemanticCache.
+
+``langchain_azure_ai.vectorstores.cache`` imports ``langchain_azure_cosmosdb``
+at module scope, but that package isn't a declared dependency of
+langchain-azure-ai and generally isn't installed in this package's dev
+environment. A minimal stub is registered in ``sys.modules`` below (before
+importing the cache module) so these tests can run without it, following the
+same pattern used for other optional SDKs in this test suite (see
+``test_speech_to_text_tool.py``).
+
+The fake vector store below intentionally mirrors the real backend's
+behaviour: every ``AzureDocumentDBVectorSearch`` created for a given
+``llm_string`` shares the *same* underlying collection -- Cosmos DB Mongo
+vCore's ``cosmosSearch`` index is defined over a field for the whole physical
+collection, not scoped by the vector store's ``index_name`` label -- and
+``similarity_search``/``delete_many`` only exclude non-matching documents
+when an explicit filter says so, exactly like the real ``cosmosSearch``
+``filter`` argument and MongoDB's ``delete_many``.
+"""
+
+from __future__ import annotations
+
+import sys
+from enum import Enum
+from types import ModuleType
+from typing import Any, Dict, List, Optional
+
+from langchain_core.documents import Document
+from langchain_core.outputs import Generation
+
+# ---------------------------------------------------------------------------
+# Stub `langchain_azure_cosmosdb` so the cache module can be imported without
+# the real package installed.
+# ---------------------------------------------------------------------------
+
+
+def _matches(doc: Dict[str, Any], flt: Dict[str, Any]) -> bool:
+    for dotted_key, expected in flt.items():
+        node: Any = doc
+        for part in dotted_key.split("."):
+            if not isinstance(node, dict):
+                return False
+            node = node.get(part)
+        if node != expected:
+            return False
+    return True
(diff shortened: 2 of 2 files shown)


santiagxf:
Thanks @cr-sbarbouche for the contribution!

PR #1070, opened 2026-09-22
fix(azure-ai): pick latest Foundry checkpoint by checkpoint id

### LINKED ISSUE #1057: bug(azure-ai): FoundryCheckpointSaver can select an older checkpoint as latest
### Description

When no `checkpoint_id` is supplied, [`FoundryCheckpointSaver.aget_tuple()`](https://github.com/langchain-ai/langchain-azure/blob/e5b12a3e7fa639374375c07dd687450fa35f44d2/libs/azure-ai/langchain_azure_ai/agents/hosting/_foundry_checkpoint_saver.py#L171-L194) selects the first result from `list_keys(order="desc", limit=1)`.

The local `FoundryStateStore` sorts by `(created_at, id)`, using second-level timestamps and hash-derived item IDs. Checkpoints written within one second can therefore be returned out of checkpoint order: a latest lookup returns an older state, while explicitly requesting the newer checkpoint ID returns the correct state.

The saver should determine checkpoint order independently of storage-item order. `alist()` and `before` filtering rely on the same assumption and should also be reviewed.

### Reproduction

1. Use `FoundryCheckpointSaver(user_isolation=False)` outside hosted mode with a temporary `AGENTSERVER_STATE_ROOT`.
2. Sequentially write two checkpoints to the same thread/namespace within one second.
3. Close/reopen the saver and call `aget_tuple()` with only `thread_id`.
4. Compare with a lookup using the second checkpoint's explicit ID. Repeat with fresh threads if needed.

Reproduced with the real clock in one process: the unqualified lookup returned `older`; the explicit lookup returned `newer`. Both checkpoints remained stored.

### Hosting scope

**The hosting-internal continuation route appears unaffected when it restores an explicit checkpoint ID.** [`ResponsesHostServer`](https://github.com/langchain-ai/langchain-azure/blob/e5b12a3e7fa639374375c07dd687450fa35f44d2/libs/azure-ai/langchain_azure_ai/agents/hosting/_responses_host.py#L651-L663) restores the saved checkpoint reference in both local and hosted execution, bypassing latest selection. A missing reference can still lead to the affected branch.

This reproduction bypasses the host; it does not demonstrate a host HITL failure. Remote State Store ordering has not been verified.

### System Info

Windows, Python 3.14.4; `langchain-azure-ai==1.2.10`, `azure-ai-agentserver-core==2.1.0`, `langgraph==1.2.11`. The same local ordering logic exists in `azure-ai-agentserver-core==2.2.0b1`.

Separate from [Azure SDK #49042](https://github.com/Azure/azure-sdk-for-python/issues/49042), which concerns cross-process lost updates.

tanvir-ux:
## Summary
- I fixed `FoundryCheckpointSaver.aget_tuple()` so that when no `checkpoint_id` is given, it picks the max LangGraph checkpoint id in the namespace instead of trusting `list_keys(order="desc", limit=1)`.
- I also updated `alist()` / `before` filtering to sort and filter by checkpoint id, because Foundry's local store orders by second-level `created_at` plus a hash-derived item id and can reverse true checkpoint order within the same second.
- I added a regression that reverses storage insertion order after writing older/newer UUID6-shaped ids and asserts latest + history stay correct.

Fixes #1057

## Test plan
- [x] `python -m pytest tests/unit_tests/agents/hosting/test_foundry_checkpoint_saver.py -q` → 11 passed
- [x] Reproduced the issue first: after `store.order.reverse()`, unqualified `aget_tuple()` returned the older id; after the fix it returns the newer one


herohua:
Thank you @tanvir-ux for your first-time contribution! Regarding the PR, could we select or sort checkpoint keys by checkpoint ID before fetching their values? Currently, even `alist(limit=2)` reads every checkpoint’s full content. Fetching values only as needed would avoid unnecessary storage requests while preserving the ordering fix.

tanvir-ux:
Thanks @herohua — good catch.

Pushed an update that lists/sorts the storage keys by checkpoint id first, then only `get_item`s the ones we actually need (latest for `aget_tuple`, or the survivors after `before` / filters / `limit` for `alist`). Added a regression so `alist(limit=2)` over five checkpoints fetches just those two bodies.

Happy to tweak further if anything still looks off.

PR #1072, opened 2026-09-22
fix(azure-ai): preserve image and file request attachments

Jiahui-Gu:
## Problem and fix

A Responses request containing text plus an image or file previously reached the LangGraph agent as text alone. Preserve ordered `input_image` / `input_file` blocks and normalize adjacent text blocks when constructing messages, including incoming tool-result history. Text-only requests retain their existing string representation.

This fixes input attachments and updates the existing Files sample to use them. Tool-output serialization, citations, arbitrary artifacts, and Invocations are outside this PR.

## Files sample cleanup

Remove the filesystem workaround from `responses/06_files`: `list_files`, `read_text_file`, path-resolution helpers, `DATA_DIR`, bundled notes, and the statements that request attachments are dropped. The sample now sends a small inline PDF via `request.json` directly to the model. Update the manual E2E check to read a verification code present only in the PDF. No custom client is added.

## Compatibility

Rich parts retain their Responses-native representation. Assistant-role attachments are preserved in graph input, but the tested gpt-4.1 Responses endpoint rejects assistant input_image/input_file content; graph preservation is not a claim that the downstream model accepts those roles. Downstream OpenAI-compatible LangChain models must use `use_responses_api=True` (as the current hosting samples do); other providers need an appropriate adapter. This does not upload files, fetch URLs, or grant access to file IDs.

## Protocol audit

Checked the request-side contract in the [Responses Create reference](https://developers.openai.com/api/reference/resources/responses/methods/create), separately from response output and SSE schemas. The generated OpenAI `ResponseInputParam` union permits image/file message parts and rich function-call-result inputs. Tests now validate the actual ChatOpenAI request input against that official type with strict validation.

The audit found three failures for image requests without `detail` (user/system/developer): Foundry accepts the omitted default, but the downstream OpenAI message schema requires the field. Normalize omitted image detail to `auto` on the copied content; preserve explicit detail and the original request. Added attachment-only coverage, not just text-plus-attachment cases.

Assistant-role blocks remain intact in graph input. Direct gpt-4.1 service checks rejected input_image/input_file in assistant messages (HTTP 400, output_text/refusal only); no upstream passthrough fix is claimed. Schema validity does not prove model format/size support, file-ID access, or deployed Foundry behavior. Cloud verification was subsequently completed; see the cloud E2E section below.

## Validation

- The original 16 regression cases fail on base `9365b59` and pass with this change. Expanded role coverage now gives **46 passing attachment tests**, including assistant-role preservation in graph input and system/developer downstream payload construction.
- Protocol audit: hosting suite **351 passed**; after adding six attachment-only cases, focused attachment tests **46 passed**. Ruff and focused mypy pass on the final change.
- Covers URL/data-URI/file-ID image inputs; URL/file-ID/inline file inputs; part order and input isolation; actual LangChain Responses payload construction; JSON/SSE requests; second-turn history with and without a checkpointer.
- Ruff, focused mypy, and `git diff --check` pass.
- Sample syntax compilation and request PDF structure checks pass. The actual sample graph and official host preserve the exact inline PDF through the downstream ChatOpenAI request (model HTTP transport mocked). The manual sample E2E runner was updated but not run. This local check used mocked model HTTP responses; the separate real-model cloud verification is documented below.
- Python 3.13; Responses SDK 2.1.0b2; LangChain Core 1.5.2 / OpenAI 1.4.1. Used the existing installed environment with this checkout on PYTHONPATH; no dependency changes.

## Before / after

Verified using Bruno against the official local ResponsesHostServer and the same deterministic LangGraph probe and request. Only the hosting source changes between runs. The request contains one image reference and one inline file: before, the agent receives **0 images / 0 files**; after, **1 image / 1 file**. The probe counts actual received blocks; it does not call a model or fetch the example image URL. This verifies the local hosting boundary, not a deployed Foundry or model-capability test.

### Before

Before: both attachments are lost

### After

After: the image and file reach the agent

[Exact local probe used for both screenshots](https://github.com/langchain-ai/langchain-azure/blob/cef0e3cd36929df835a91cba4882ec0c7fff196a/.github/pr-assets/1072/attachment-probe.py). Screenshots and probe are stored on a separate assets branch and are not part of this PR diff.

## Cloud validation: request attachments

Validated this PR's input path through a real Foundry hosted agent and gpt-4.1, using the same inline PNG/PDF and prompt. Codes occur only inside attachments. JSON and SSE both show:

| Input check | Before | After |
|---|---|---|
| Attachments received by agent | 0 | 2; exact normalized block hashes match |
| Input attachments in next-turn history | 0 | 2 preserved |
| Model reads image/PDF codes | MISSING | MAPLE-7319 and ORCHID-4827 |

**Before — real cloud response:**

Cloud input before: model receives no attachments

**After — real cloud response:**

Cloud input after: model reads both attachment codes

[Input-only JSON evidence](https://github.com/langchain-ai/langchain-azure/blob/00f67747df69dc466e2218a3760470c881557ca0/.github/pr-assets/1072/pr-1072-cloud-evidence.json). Screenshots are actual Bruno requests to Foundry. The common test deployment used baseline v1 and a combined-fix v2, with explicit version-bound sessions; this section reports only request-input behavior attributable to this PR (ab63b71). No URL/file-ID authorization, other formats/models, or cloud checkpoint claim.


samples/hosting/langgraph-hosted-agents/responses/06_files/README.md
```diff
@@ -1,93 +1,75 @@
...
 
-The upstream Agent Framework sample uploads files into a hosted-agent
-session and references them by `file_id` in the Responses input. Today
-the langchain Responses host only forwards **text** content blocks (see
-the hosting converter at
-[`libs/azure-ai/langchain_azure_ai/agents/hosting/_converters/_request.py`](../../../../libs/azure-ai/langchain_azure_ai/agents/hosting/_converters/_request.py))
-— `input_file` / `input_image` items are dropped. Until the hosting
-layer learns to passthrough those items, this sample exposes files to
-the agent via the filesystem inside the container rather than as
-request attachments. The user-facing pattern is the same: "the agent
```
Jiahui-Gu, line 25: Reviewer note: this deleted paragraph explicitly documented the existing bug: `input_file` / `input_image` items were dropped, so the sample used container filesystem tools instead of request attachments. The request-converter fix in this PR removes that limitation. The associated sample cleanup removes that workaround (`list_files`, `read_text_file`, `DATA_DIR`, and bundled notes) and replaces it with a direct PDF request plus an updated manual E2E check. Please review these deletions together with the converter change; they are the removal of a documented workaround, not an unrelated sample rewrite.

libs/azure-ai/langchain_azure_ai/agents/hosting/_converters/_request.py
```diff
@@ -131,13 +132,13 @@ def _item_call_id(item: Any) -> str | None:
 
 def _item_to_message(item: Any) -> AnyMessage | None:
     if _is_message(item):
-        text = _content_to_text(item["content"])
+        content = _content_to_message_content(item["content"])
```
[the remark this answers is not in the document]
Jiahui-Gu: Confirmed with langchain-openai 1.4.1: an attachment-only assistant AIMessage produces an empty downstream input, and mixed assistant text/file content retains only text. However, items_to_messages already preserves the complete attachment blocks and assistant role; the loss happens later in langchain-openai._construct_responses_api_input.

In cceef2f2e7d9d47bccd12f9160c40d62de310d7a, added 12 assistant-role cases (attachment-only and mixed text for all six reference forms) verifying host preservation, role, and copy isolation. Also expanded downstream payload coverage to system/developer inputs and documented the assistant serialization limitation in the sample and PR description. All 40 attachment tests, Ruff, and focused mypy pass.

Disposition: downstream serializer support is outside this user-input hosting fix. Relabeling the AIMessage as user or bypassing the provider serializer would change conversation semantics; this PR does neither. Resolving this thread with the limitation explicitly documented, not claiming assistant attachments are now forwarded by langchain-openai.

samples/hosting/langgraph-hosted-agents/responses/06_files/main.py
```diff
@@ -1,121 +1,26 @@
-"""Sample 06 - Responses API + filesystem tools.
```
herohua, line 1: I guess this sample wants to demo on filesystem tools. Invoking by image/file request attachments is not its original intention.

Jiahui-Gu: ref: https://github.com/langchain-ai/langchain-azure/pull/1072#discussion_r4069073805

Wixee (approved):

samples/hosting/langgraph-hosted-agents/responses/06_files/main.py
```diff
@@ -1,121 +1,26 @@
...
-
-    az login
-    cp .env.example .env  # then edit the values
-    python main.py
-
-Then in another terminal::
-
-    curl -X POST http://127.0.0.1:8088/responses -H 'Content-Type: application/json' -d '{"input":"List the files available to you, then summarize notes.txt.","model":"gpt-4o"}'
+Run with FOUNDRY_PROJECT_ENDPOINT and AZURE_AI_MODEL_DEPLOYMENT_NAME set.
+Send request.json to /responses to read the inline PDF directly with the model.
```
herohua: Could we retain the short setup/run guide in this module docstring, consistent with samples 01_basic and 02_tools? The required FOUNDRY_PROJECT_ENDPOINT, optional AZURE_AI_MODEL_DEPLOYMENT_NAME (default gpt-4o), PORT (default 8088), and az login / .env / startup steps still apply. Update the request example to send request.json and explicitly note that the chosen deployment must support multimodal input, the Responses API, and PDF input for this example. The obsolete DATA_DIR and filesystem-tool instructions can stay removed.

samples/hosting/langgraph-hosted-agents/responses/06_files/README.md
```diff
@@ -1,93 +1,81 @@
...
-can read files at runtime"; the wiring is different.
+Follow the [local setup instructions](../../README.md#running-the-agent-host-locally).
+Use a model deployment that supports the attachment type and the Responses API.
```
herohua: Could we preserve the brief Model Integration and Agent Hosting explanations used by the neighboring samples? DefaultAzureCredential, AIProjectClient, ChatOpenAI, and ResponsesHostServer still form the same flow; only the attachment handling changes. Please also explicitly require a multimodal deployment supporting the Responses API and PDF input for request.json, and clarify that file_data contains an actual PDF encoded as Base64, sent as an input_file block rather than ordinary chat text. This would help readers understand why pasting the Base64 string into a text-only Playground field does not test the attachment path.

samples/hosting/langgraph-hosted-agents/responses/06_files/main.py
Jiahui-Gu: Restored the concise module setup/run guide in a6c82c8: required project endpoint, optional deployment and port defaults, az login/.env/startup steps, and the request.json curl command. It explicitly requires multimodal Responses/PDF support. DATA_DIR and filesystem workaround instructions remain removed. Syntax compilation and diff checks passed.

samples/hosting/langgraph-hosted-agents/responses/06_files/README.md
Jiahui-Gu: Added brief Model integration and Agent hosting sections in a6c82c8, retaining the credential -> project endpoint -> ChatOpenAI -> ResponsesHostServer explanation. Also explicitly states that request.json carries an actual PDF encoded as Base64 in input_file.file_data; pasting Base64 into a text Playground is not the attachment path. Corrected the assistant-role note to reflect the observed service rejection rather than asserting an upstream serializer fix is required.

samples/hosting/langgraph-hosted-agents/responses/06_files/README.md
```diff
@@ -1,93 +1,96 @@
...
+`ChatOpenAI` model forwards them using `use_responses_api=True`.
 
-- `list_files(subpath="")` — list the contents of a directory below the
-  configured data root.
-- `read_text_file(file_path)` — read a UTF-8 text file from the data
-  root (capped at 64 KiB).
+Send attachments with `role: "user"` as in this sample. Do not infer that
+assistant-role attachments are supported from the host preserving graph input:
+the tested gpt-4.1 Responses endpoint rejects assistant `input_image` /
+`input_file` content. Changing the message role would change conversation semantics.
```
herohua: This paragraph is confusing in the context of this sample: the client sends a question and attachments in a `user` message, and the hosted agent returns an `assistant` reply. Introducing assistant-role input attachments here makes it sound as though the client needs to send them as part of the normal workflow. Although clients can replay assistant messages as conversation history, that is a separate concern, and replaying history does not change the original user attachments into assistant attachments.

Could we replace this paragraph with:

> Send your question and attachments in a `user` message, as shown in `request.json`.

The assistant-role compatibility limitation and the specific gpt-4.1 test result can stay in the PR validation notes or test documentation, rather than the sample introduction.

Jiahui-Gu: Agreed. Replaced the paragraph with the suggested user-message instruction in 5df5f4f. The assistant-role/service compatibility details remain in the PR validation notes rather than the sample introduction. This is documentation-only; git diff --check passed.

herohua (approved):

